PR #1, opened 2023-12-08
[LMB-37] add simple route to render basic form definition

Rahien:


bfidlers (commented):
Can't really get it to work.
Whenever I try to load the page I get the following error message, not sure if I need to load a form somewhere or something like that: 
```
Error while processing route: form.new JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

bfidlers:
Still get the same error as mentioned above.

bfidlers (approved):

PR #3, opened 2023-12-11
[LMB-25] allow creating form instances

Rahien:
This PR builds on top of https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/1, please review that one first!

This adds the possibility to send the create call in the form/new route. This call simply sends the form source graph to the backend.


bfidlers (approved):
Seems good and everything seems to be working. 
Didn't test if it actually gets written to the database, but the form-content service seems to receive it and execute the query, so should be fine I guess.

PR #5, opened 2023-12-15
Fix deprecations

bfidlers:
LMB-51
There are still a few deprecations left, for example the following two:
[AuInput] The `@value` argument is deprecated. Use the `value` attribute and the `{{on}}` modifier instead.
[AuInput] The masking feature is deprecated. Use the `{{au-inputmask}}` modifier instead.

The masking feature should be solvable, but I couldn't get my head around it, the @value one is a bit more widespread and difficult to solve in a decent amount of time.

There also is some old syntax that the linter is complaining about, but it would turn into a much bigger effort to fix these than it is actually worth. (Old component extensions etc.).

### DIFF
--- a/app/components/shared/persoon/create-persoon.hbs
+++ b/app/components/shared/persoon/create-persoon.hbs
@@ -139,22 +139,20 @@
         </AuLabel>
         <div class="au-o-grid au-o-grid--tiny">
           <div class="au-o-grid__item au-u-1-2 au-u-1-4@medium">
-            <AuControlRadio
-              @label="Vrouw"
-              checked={{this.isFemale}}
+            <AuRadio
+              @checked={{this.isFemale}}
               @value={{this.female}}
               @onChange={{this.setGender}}
               @name="gender"
-            />
+            >Vrouw</AuRadio>
           </div>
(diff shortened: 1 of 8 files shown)


app/components/shared/persoon/create-persoon.hbs
```diff
@@ -139,22 +139,20 @@
...
-            <AuControlRadio
-              @label="Vrouw"
-              checked={{this.isFemale}}
+            <AuRadio
+              @checked={{this.isFemale}}
               @value={{this.female}}
               @onChange={{this.setGender}}
               @name="gender"
-            />
+            >Vrouw</AuRadio>
```
Rahien, line 147: seems limited :grimacing:

app/templates/mandatenbeheer/fracties.hbs
```diff
@@ -85,19 +85,19 @@
               </thead>
               <tbody>
                 {{#if this.newFractie}}
-                  {{mandatenbeheer/fractie-table-row
```
Rahien, line 88: :sauropod:

Rahien (approved):

PR #6, opened 2023-12-19
fetch and render form instance

Rahien:
This passes the uri of the instance to be created to the form service and expects the service to return the uuid of the instance. It then transitions to the (new) instance route, rendering the same form instance, fetched from the server in case additional content is added server-side that we don't know about. this isn't the case right now but allows for future extensions.


bfidlers (approved):
Looks good and seems to work as expected. In the future there should be better visual feedback that a form has actually been filled in, but those are future considerations.

PR #7, opened 2023-12-19
update form instance

Rahien:
Allow updating a form instance. Builds on fetching a form instance: https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/6

image


bfidlers (approved):
Looks good and seems to work, but same remark is in the other review, in the future we need to think about how to represent this, to make it clear that you are editing a form and not just filling in a new one or something like that. But technically it certainly is ok and works.

PR #14, opened 2024-01-02
Resource clean up

bfidlers:
Re-added some resources or relations to the frontend to better reflect the original domain model.


app/models/werkingsgebied.js
ruben-bottu, line 1: Are the extra local vars for niveau and naam necessary?

ruben-bottu (approved):

PR #23, opened 2024-01-19
allow using hidden fields in form to track behind-the-scenes properties

Rahien:
This PR refactors the forms so that there are components for managing form instances. It then also creates a fracties-new route where behind-the-scenes data for fracties is passed in (bestuurseenheid and bestuursorganen) like it is done in the hard-coded legacy view of fracties


app/components/form/instance-table.js
```diff
@@ -0,0 +1,61 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+
+import { inject as service } from '@ember/service';
+
+export default class InstanceTableComponent extends Component {
+  @service store;
+
+  @tracked formInfo = null;
```
ruben-bottu, line 10: I'd rather not have null in a dynamically typed language, but I guess it's inevitable

app/components/form/instance-table.js
```diff
@@ -0,0 +1,61 @@
...
+    if (this.args.editRoute) {
+      return this.args.editRoute;
+    }
+    return 'form.instance';
+  }
+
+  @action
+  async removeInstance(instanceId) {
+    const result = await fetch(
+      `/form-content/${this.formInfo.form.definition.id}/instances/${instanceId}`,
```
ruben-bottu, line 26: Maybe the route `/form-content` should be in a .env file?
Generally could be safer to have a helper method that puts the "/" for you, so you don't accidentally break the URI

app/components/form/instance-table.js
```diff
@@ -0,0 +1,61 @@
...
+    if (this.args.onRemoveInstance) {
+      this.args.onRemoveInstance(instanceId);
+    }
+  }
+
+  @action
+  async onInit() {
+    const form = this.args.form;
+    const id = form.definition.id;
+    const response = await fetch(`/form-content/${id}/instances`);
```
ruben-bottu, line 49: Same as above

app/components/form/instance.hbs
```diff
@@ -0,0 +1,47 @@
...
+              @form={{this.formInfo.form}}
+              @show={{false}}
+              @graphs={{this.formInfo.graphs}}
+              @sourceNode={{this.formInfo.sourceNode}}
+              @formStore={{this.formInfo.formStore}}
+              @forceShowErrors={{true}}
+              @useNewListingLayout={{true}}
+              @level={{2}}
+              class="au-u-max-width-medium"
+            />
```
ruben-bottu, line 20: Makes me wonder again if we should create a generic component to avoid code duplication

app/components/form/instance.js
```diff
@@ -0,0 +1,136 @@
...
+  get initialized() {
+    return this.formInfo !== null;
+  }
+
+  get isSaving() {
+    return this.save.isRunning;
+  }
+
+  @keepLatestTask
+  *save() {
```
ruben-bottu, line 29: Maybe this method is a bit long?

ruben-bottu (approved):

app/components/form/instance-table.js
Rahien: i'd avoid too many layers of indirection for now, the app is already quite complicated this would add to it for little gain imo

app/components/form/instance-table.js
Rahien: a way around this would be to add a placeholder value while loading but i think this doesn't add too much (and would complicate the initialized getter)

app/components/form/instance.hbs
Rahien: normally these components should guarantee that there are few instances of rdfform being created, i.e. you should use these components as the generic components. Adding another layer of indirection would complicate it too much again imo

app/components/form/instance.js
Rahien: feels acceptable tbh

PR #24, opened 2024-01-22
Type selector fetching

bfidlers:
Example form is included in branch type-selector of the app-lokaal-mandatenbeheer

This PR builds further upon #22, now the data is pulled from resources and the correct predicates are used form the form field. 
One problem is that for the multi-instance-selector, multiple calls are needed to the backend to fetch filled in values. This is an issue with resources, we should fix this in the future.

This PR should close both LMB-42 and LMB-43.


app/components/rdf-input-fields/instance-multi-selector.js
```diff
@@ -0,0 +1,143 @@
...
+  async load() {
+    await this.loadOptions();
+    await this.loadProvidedValue();
+  }
+
+  async loadOptions() {
+    const instanceLabelProperty = this.getFormProperty('instanceLabelProperty');
+    const instanceApiUrl = this.getFormProperty('instanceApiUrl');
+
+    const pageSize = 5;
```
Rahien, line 37: should this be higher? probably done for testing of pagination

app/components/rdf-input-fields/instance-multi-selector.js
```diff
@@ -0,0 +1,143 @@
...
+
+export default class RdfInstanceMultiSelectorComponent extends InputFieldComponent {
+  inputId = 'select-' + guidFor(this);
+
+  @tracked selected = null;
+  @tracked options = [];
+  @tracked searchEnabled = true;
+
+  get subset() {
+    return this.options.slice(0, 50);
```
Rahien, line 20: why slice options if we already have pagination

app/components/rdf-input-fields/instance-multi-selector.js
```diff
@@ -0,0 +1,143 @@
...
+      const subject = new NamedNode(m.attributes['uri']);
+      return { subject: subject, label: m.attributes[instanceLabelProperty] };
+    });
+    return options;
+  }
+
+  // TODO we want to do this with one call, something like the following unfortunately this doesn't work at this moment.
+  // const url = `${instanceApiUrl}?filter[:uri:]=uri1,uri2`;
+  // const url = `${instanceApiUrl}?filter[:or:][:uri:]=uri1&[:or:][:uri:]=uri2`;
+  async uriSearch(term) {
```
Rahien, line 131: suggestion: rename term to targetUri? term to me sounds like you're looking up by label which you shouldn't and aren't :+1:

app/components/rdf-input-fields/instance-multi-selector.js
```diff
@@ -0,0 +1,143 @@
...
+        })
+      );
+      this.selected = options
+        .map((m) => m[0])
+        .filter((opt) => matches.find((m) => m.equals(opt.subject)));
+    }
+  }
+
+  @action
+  updateSelection(options) {
```
Rahien, line 62: suggestion: rename to newSelection to make the difference with this.options more clear?

Rahien (approved):
approved, but left some suggestions

app/components/rdf-input-fields/instance-multi-selector.js
bfidlers: Yeah this is purely for testing, what would you suggest? 10?

PR #26, opened 2024-01-23
Rebased form navigation

bfidlers:


app/router.js
```diff
@@ -61,11 +61,13 @@ Router.map(function () {
     });
   });
 
-  this.route('form', { path: '/form/:id' }, function () {
-    this.route('new');
-    this.route('instances');
-    this.route('instance', { path: '/instance/:instance_id' });
-    this.route('edit');
+  this.route('formbeheer', function () {
+    this.route('form', { path: '/form/:id' }, function () {
```
Rahien, line 65: doesn't this make the path /formbeheer/form/id? why not /formbeheer/id?

Rahien (approved):

app/router.js
bfidlers: I think this is still needed because the form still has a route to fetch the form-definition etc.

PR #29, opened 2024-01-24
format triples preview a little nicer

Rahien:
image


bfidlers (approved):
Small fix, big improvement 👍

bfidlers:
Might also work as a more general component, but is fine for now.

PR #30, opened 2024-01-26
Introduce some extra constants to prevent prevent hardcoded URI's, magic variables etc.

bfidlers:


app/components/form/instance-table.js
```diff
@@ -31,9 +32,7 @@ export default class InstanceTableComponent extends Component {
       `/form-content/${this.formInfo.form.definition.id}/instances/${instanceId}`,
       {
         method: 'DELETE',
-        headers: {
-          'Content-Type': 'application/vnd.api+json',
-        },
+        headers: CONTENT_HEADER,
```
Rahien, line 35: JSON_API_CONTENT_HEADER?

app/components/leidinggevendenbeheer/functionaris-form.js
```diff
@@ -2,6 +2,7 @@ import Component from '@glimmer/component';
 import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
+import { FUNCTIONARIS_STATUS_CODE_AANGESTELD_URI } from 'frontend-lmb/utils/constants';
```
Rahien, line 5: thinking: separate file called well-known-uris? so you don't have to put _uri at the end

app/components/rdf-input-fields/instance-multi-selector.js
```diff
@@ -32,11 +34,10 @@ export default class RdfInstanceMultiSelectorComponent extends InputFieldCompone
...
     const instanceApiUrl = this.getFormProperty('instanceApiUrl');
 
-    const response = await fetch(`${instanceApiUrl}?page[size]=${PAGESIZE}`, {
-      headers: {
-        Accept: 'application/vnd.api+json',
-      },
-    });
+    const response = await fetch(
+      `${instanceApiUrl}?page[size]=${PAGESIZE}`,
+      ACCEPT_HEADER
```
Rahien, line 39: i feel like this may be too much, i would just use the value as the constant

app/components/form/instance-table.js
Rahien: actually would just put the value for content type in the constant so you can extend headers if you want (without the hassle of cloning etcetc)

app/components/rdf-input-fields/instance-multi-selector.js
Rahien: since all headers in this file are the same you could make that a local constant if you'd want, not sure if necessary though

app/components/shared/persoon/create-persoon.js
```diff
@@ -12,9 +12,8 @@ import {
   isValidRijksregisternummer,
 } from 'frontend-lmb/utils/rijksregisternummer';
 import moment from 'moment';
+import { FEMALE_ID, MALE_ID } from 'frontend-lmb/utils/constants';
```
Rahien, line 15: thinking: well-known-ids file? that would keep constants a little more organized too

app/components/leidinggevendenbeheer/functionaris-form.js
Rahien: or a hash with a name like that in the constants could also work

Rahien (changes requested):

app/components/rdf-input-fields/instance-multi-selector.js
bfidlers: This pagesize value was already there and a decision that was made when making this component, since it is a constant defined in this file for calls made from this component I don't really see a problem with this and personally I would prefer this.

PR #34, opened 2024-01-31
Mandatenbeheer v2

bfidlers:
LMB-72 Wrappers for mandatenbeheer.


app/components/mandatenbeheer/bestuursperioden-selector.hbs
Rahien, line 1: I see you create new components that mirror the legacy ones. This is okay if you need to change them substantially, but if you don't i guess we are okay just moving the legacy ones back into the main component folder, no?

Rahien (approved):

PR #35, opened 2024-02-01
Full Readme Update

ruben-bottu:


README.md
```diff
@@ -32,12 +54,16 @@ The [ember-proxy-service](https://github.com/mu-semtech/ember-proxy-service#conf
 
 ### Feature flags
 
+> TODO This functionality is not supported yet.
```
Rahien, line 57: it seems we removed the feature flags functionality. not sure if we should have as it is a very useful thing to have when the app is more mature. anyway either we add it back in or we remove this section. see: https://github.com/lblod/frontend-organization-portal?tab=readme-ov-file#feature-flags

For now i would remove. Will create a ticket to add feature flags back in with this link for reference

Rahien (changes requested):

PR #38, opened 2024-02-02
add address-selector rdf-input component

Rahien:
this adds a selector components for addresses, new address records are created for every change, so we will need a service that cleans unconnected addresses, i will create a ticket for this.

The component can be tested by adding this snippet to the test form:

```
ext:testAdresSelector
    a form:Field;
    form:displayType displayTypes:addressSelector;
    sh:group ext:testPG;
    sh:name "testAdresSelector";
    sh:order 5;
    sh:path ext:testAdres.
```

And adding this field to the form:includes list

The result looks something like this (searching for the address of the flemish parliament):

image


app/components/extra/adressenregister-selector.js
bfidlers, line 1: Maybe use the search timeout constant for this? this constant is already included in master.

app/components/rdf-input-fields/address-selector.hbs
```diff
@@ -0,0 +1,8 @@
+{{#if this.initialized}}
+  <Extra::AdressenregisterSelector
+    @address={{this.address}}
+    @onChange={{this.updateAddress}}
+  />
+{{else}}
+  Loading...
```
bfidlers, line 7: Maybe we should use the page-loader component here?

app/components/extra/adressenregister-selector.hbs
bfidlers, line 1: These components should be in another folder, maybe address or something like that.
Also there is some special syntax here in the yield, this needs to be reviewed if this can be made more readable and less complex.

bfidlers:
Still some remarks, mainly about legacy code, but otherwise looks good and seems to work fine.

bfidlers (changes requested):

app/components/rdf-input-fields/address-selector.hbs
Rahien: yes!! there is already a story for this! I think we should use skeleton loaders

app/components/extra/adressenregister-selector.js
Rahien: yup! will do! thanks for adding that!

PR #39, opened 2024-02-05
add mandataris view page

bfidlers:


app/components/mandatenbeheer/mandataris/mandataris-view.js
```diff
@@ -0,0 +1,39 @@
...
+  @service() store;
+
+  @tracked mandatarissen;
+
+  constructor() {
+    super(...arguments);
+    this.initComponentProperties();
+  }
+
+  async initComponentProperties() {
```
bfidlers, line 15: Might be better to move this to the route.

Rahien (approved):

PR #40, opened 2024-02-05
add person selector/creator

Rahien:
image
image
image
image
image

includes creating a person (in the same way that this was done before in Loket)


app/components/person/create.hbs
bfidlers, line 1: I think this should become a form as well, shouldn't we flag this in a todo component or something like that? And maybe put this back to legacy, because in that case this component should disappear in the near future.

app/components/person/search-form.hbs
bfidlers, line 1: This component is a little too big for my liking, might be worth it to review this later and simplify this or maybe split up some logic in different components that could be reused in other search components or something like that.

bfidlers (commented):
Few remarks:
- Ideally legacy templates should never be changed, I get why these components are moved from legacy to a non-legacy folder, since these haven't really been refactored, the problem however is if you do change some logic in these components your legacy routes break or you have to refactor these again ...
- Layout wise I think the rdf-input component could be better, now there is a bin icon even if no person has been selected, the button is aligned to the right which doesn't really make sense and "bewerk" seems a bit weird if you want to add a person, but UI/UX might be something we want to fix in a later stage
- The legacy components that are moved here throw a lot of deprecation warnings, I think some of these might be a bit finicky to fix, but it should at least be checked

app/components/person/create.hbs
Rahien: good point! will move it back. there is actually a ticket for this in the next sprint

app/components/person/list-item.js
```diff
@@ -0,0 +1,18 @@
...
+export default class SharedPersoonPersoonSearchFormComponent extends Component {
+  get personText() {
+    const person = this.args.person;
+    if (!person) {
+      return 'Geen persoon geselecteerd';
+    }
+    return `${person.gebruikteVoornaam} ${person.achternaam}`;
+  }
+
+  get personRrn() {
```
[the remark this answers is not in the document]
Rahien: was for an idea i had to show the rrn on hover but ended up not doing it. will remove

app/components/person/search-form.hbs
Rahien: hmmm i agree that this is a large component but splitting it up doesn't feel like the purpose of this story. i can make a separate ticket for it?

Rahien:
> * Layout wise I think the rdf-input component could be better, now there is a bin icon even if no person has been selected, the button is aligned to the right which doesn't really make sense and "bewerk" seems a bit weird if you want to add a person, but UI/UX might be something we want to fix in a later stage

This is completely true. We will get designer support starting from thursday, which is why I didn't bother too much with design/ux for now.

bfidlers (approved):
With the given remarks in mind for future reference, this seems okay, and also works btw.

PR #42, opened 2024-02-05
Add fraction form

ruben-bottu:


app/utils/get-form.js
```diff
@@ -0,0 +1,43 @@
...
+    throw error;
+  }
+
+  const form = await response.json();
+  if (!form.formTtl) throw new Error('Received form without formTtl');
+
+  return form;
+};
+
+// Split into definition and prefix because this structure is expected in instance components
```
Rahien, line 15: as the prefix is now part of the definition, wouldn't it make more sense to change the instance components?

Rahien (approved):

app/utils/toasts.js
ruben-bottu, line 1: As Ben stated, this might be a bit overkill. Any opinions @Rahien ?

app/utils/loadForm.js
```diff
@@ -0,0 +1,12 @@
+const loadForm = async (
```
Rahien, line 1: now that this isn't local to anything i think it could do with a rename, maybe loadFormAsStore?

app/utils/well-known-ids.js
```diff
@@ -1,2 +1,4 @@
 export const MALE_ID = '5ab0e9b8a3b2ca7c5e000028';
 export const FEMALE_ID = '5ab0e9b8a3b2ca7c5e000029';
+export const FRACTIE_FORM_ID = 'fractie';
+export const FORM_DEFINITION_MODEL = 'form-definition';
```
Rahien, line 4: this may be driving it a little far... model names are normally just used as a string in ember

Rahien (approved):

PR #43, opened 2024-02-06
remove fracties-new and bring legacy routes inline with form definition model change

Rahien:


bfidlers:
Also fracties-new is still in the controller and in the routes folder

PR #48, opened 2024-02-08
fix form content not loaded in time

Rahien:
image

also fixes the scrolling in the mandataris edit view


app/components/form/instance.js
ruben-bottu, line 1: Slightly hacky I'd say, but it works

ruben-bottu (approved):

PR #49, opened 2024-02-08
fractie relation

Rahien:
image

the fractie component needs context to show the correct set of fractions (i.e. the ones related to the currently selected bestuursorgaan)
Right now, it just takes the first bestuursorgaan in the list of current bestuursorganen, but since the fracties are linked to all of those that's fine for now. We will change to a setup where a single bestuursorgaan is selected next sprint

The bestuursorgaan is passed in to the meta graph of the form

There is also a business rule that dictates that the membership should have the same timeframe as the mandataris' start to end period. This is enforced after save of the form.

To use this component, add the following ttl to the form:

```
ext:fractieF
    a form:Field;
    form:displayType displayTypes:mandatarisFractieSelector;
    sh:group ext:editMandatarisEindePG;
    sh:name "Fractie";
    sh:order 3;
    sh:path org:hasMembership.

```


app/components/rdf-input-fields/mandataris-fractie-selector.hbs
bfidlers, line 1: Should probably be "mandataris-fractie-selector"

bfidlers:
Can't really use this as is, because when using this in a new form it misses the applicationContext and it isn't included in the mandate form yet. So I would suggest leaving this for now and revisiting this next week after the demo.

app/routes/mandatenbeheer/mandatarissen/edit.js
```diff
@@ -26,6 +26,9 @@ export default class MandatenbeheerMandatarissenEditRoute extends Route {
     return RSVP.hash({
       bestuurseenheid: parentModel.bestuurseenheid,
       bestuursorganen: parentModel.bestuursorganen,
+      // TODO temporary workaround, we will switch to an application layout where only a  single
+      // bestuursorgaan is selected
+      bestuursorgaan: parentModel.bestuursorganen[0],
```
bfidlers, line 31: I think it is a bit weird to pass bestuurorganen and bestuursorgaan both, but since this will change in the future I can live with it.

app/utils/bestuursorgaan-meta-ttl.js
```diff
@@ -0,0 +1,14 @@
+export const getBestuursorgaanMetaTtl = (bestuursorgaan) => {
+  if (!bestuursorgaan) {
+    return;
+  }
+  const bestuursorgaanUri = bestuursorgaan.uri;
+
+  return `
+    @prefix mandaat: <http://data.vlaanderen.be/ns/mandaat#> .
+    @prefix ext: <http://mu.semte.ch/vocabularies/ext/> .
+    @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
```
bfidlers, line 10: Not sure why you need mandaat and xsd as prefix if they are not used....

app/components/form/new-instance.js
```diff
@@ -73,7 +73,10 @@ export default class NewInstanceComponent extends Component {
     notifyFormSavedSuccessfully(this.toaster);
 
     if (this.args.onCreate) {
-      this.args.onCreate(id);
+      this.args.onCreate({
+        instanceTtl: triples,
+        instanceId: id,
+      });
```
bfidlers, line 79: This is a bit weird, you add arguments that are never used in this version.. (instanceTtl is never used I think).

app/components/form/instance.js
```diff
@@ -76,7 +76,11 @@ export default class InstanceComponent extends Component {
     notifyFormSavedSuccessfully(this.toaster);
 
     if (this.args.onSave) {
-      this.args.onSave(body);
+      this.args.onSave({
+        instanceId,
+        instanceTtl: triples,
+        response: body,
+      });
```
bfidlers, line 83: I'm not sure if all these arguments are used, but I would need to recheck this.

bfidlers (approved):
Few remarks, but overall seems good and works.

app/routes/mandatenbeheer/mandatarissen/edit.js
Rahien: agreed!

app/utils/bestuursorgaan-meta-ttl.js
Rahien: will remove

app/components/form/new-instance.js
Rahien: it's done because the onSave for mandataris needs it. It uses the instanceTtl to fetch the period of the mandatee and set their fraction membership to the same period every time (on create and update)

PR #50, opened 2024-02-08
add mandaat selector component

Rahien:
because the set of mandaten is limited to the ones of the current bestuursorgaan AND because the name of the mandaat is actually the label of its codelist item, we need a custom component to select the mandate of a mandataris

image

This PR builds on https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/49 and merges into it if that hasn't been merged to master yet.

can be added to a form by adding

```
ext:mandaatF
    a form:Field;
    form:displayType displayTypes:mandatarisMandaatSelector;
    sh:group ext:editMandatarisEindePG;
    sh:name "Mandaat";
    sh:order 5;
    sh:path org:holds.
```


app/utils/register-form-fields.js
bfidlers, line 1: Typo: Should be RDFMandatarisMandaatSelectorComponent

app/components/rdf-input-fields/mandataris-mandaat-selector.js
```diff
@@ -0,0 +1,69 @@
...
+    this.load();
+  }
+
+  async load() {
+    await this.loadProvidedValue();
+    await this.loadBestuursorgaan();
+    this.initialized = true;
+  }
+
+  async loadBestuursorgaan() {
```
bfidlers, line 31: This is code is already present in mandataris-fractie-selector, might by worth it to extract this, since we will probably need it again somewhere.

bfidlers: or maybe use some inheritance?

app/components/mandatenbeheer/mandaat-bestuursorganen-selector.js
```diff
@@ -0,0 +1,37 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { inject as service } from '@ember/service';
+
+export default class MandaatBestuursorganenSelector extends Component {
+  @service store;
+
+  @tracked mandaat = null;
+  @tracked mandaten = null;
```
bfidlers, line 10: mandaat - mandaten, might be a bit confusing, might be a good idea to indicate the difference with selected or with options?

bfidlers (approved):
The selector does not show all possible mandates that are shown in legacy. This is because only one bestuursorgaan-in-de-tijd is selected. This problem should hopefully solve itself in the future, but it is crucial to keep this in mind.

Made some minor remarks that could improve this PR, but it is acceptable as is.

app/utils/form-context/bestuursorgaan-meta-ttl.js
bfidlers, line 1: Did you call it form-context on purpose our should this be form-content?

bfidlers (approved):
Linter fails but otherwise okay.

PR #52, opened 2024-02-12
some improvements for demo purposes

bfidlers:
Some small improvements for demo purposes:
- Forms now have the option to include a cancel button
- If you select a mandate without a fractie, the frontend doesn't crash anymore
- Some info is toggled on and off when correcting mistakes on a mandate

### DIFF
--- a/app/components/form/instance.hbs
+++ b/app/components/form/instance.hbs
@@ -18,11 +18,30 @@
             class="au-u-max-width-medium"
           />
           {{yield}}
-
         </section>
-        <AuButton {{on "click" this.saveInstance}} @disabled={{this.isSaving}}>
-          Pas aan
-        </AuButton>
+
+        <AuToolbar as |Group|>
+          <Group>
+            <AuButtonGroup>
+              <AuButton
+                {{on "click" this.saveInstance}}
+                @disabled={{this.isSaving}}
+              >
+                Pas aan
+              </AuButton>
+              {{#if @onCancel}}
+                <AuButton
+                  {{on "click" this.cancel}}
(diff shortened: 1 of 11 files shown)


app/components/form/instance.hbs
```diff
@@ -18,11 +18,30 @@
...
+                {{on "click" this.saveInstance}}
+                @disabled={{this.isSaving}}
+              >
+                Pas aan
+              </AuButton>
+              {{#if @onCancel}}
+                <AuButton
+                  {{on "click" this.cancel}}
+                  @skin="secondary"
+                  {{!-- @disabled={{not @cancel}} --}}
```
Rahien, line 36: would remove this

app/components/mandatenbeheer/mandataris/mandataris-edit-prompt.hbs
```diff
@@ -1,44 +1,48 @@
...
-        verandert:
-      </p>
-      <ul>
-        <li>beëindig dit mandaat door de einddatum in te vullen;</li>
-        <li><strong>start een nieuw mandaat</strong>
-          met de gewijzigde situatie.</li>
-      </ul>
-      <AuButton {{on "click" this.terminate}}>Wijzig einde mandaat</AuButton>
-    </AuContent>
+{{#if this.notEditing}}
```
Rahien, line 1: can be `{{#if (not this.editMode)}}` so you don't need the extra getter that's just a negation

Rahien (approved):
Approved but left some comments that could make it even more clean!

PR #53, opened 2024-02-13
LMB-129 Table refactor

bfidlers:
Three tables have been properly refactored:
- mandataris table
- functionaris table
- fractie table
They now all use the AuDataTable and have sorting and pagination. 
In the future it would be nice to combine this in one component however.

The instance table also received a refactor, but does not allow for sorting etc, this is a problem that should be looked into at a later time if needed (this needs changes in the form-content service as well).


app/components/form/instance-table.hbs
```diff
@@ -1,41 +1,46 @@
...
-        Form Instances</AuHeading>
-    </Group>
-    <Group>
-      <AuButton {{on "click" @onCreate}}>Voeg form instance toe</AuButton>
-    </Group>
-  </AuToolbar>
+<AuToolbar @size="large" class="au-u-padding-bottom-none" as |Group|>
+  <Group>
+    <AuHeading @skin="2">Beheer
+      {{this.formInfo.form.definition.id}}
```
ruben-bottu, line 4: Shouldn't this be this.formInfo.formDefinition.id?

app/components/mandatenbeheer/mandataris/mandataris-table.hbs
```diff
@@ -0,0 +1,79 @@
...
+      <th></th>
+    </c.header>
+
+    <c.body as |row|>
+      <td>
+        {{#each
+          (get @mandatarisBestuursorganen row.id)
+          as |bestuursorgaan index|
+        }}
+          {{#if (gt index 0)}}<br />{{/if}}
```
ruben-bottu, line 52: This seems a bit hacky, but ok

ruben-bottu (approved):

Rahien (approved):
very pretty!

app/components/form/instance-table.hbs
bfidlers: good catch, seems like it was broken before as well.

app/components/mandatenbeheer/mandataris/mandataris-table.hbs
bfidlers: Not my code, but should resolve itself in the future, with the new wireframes.

PR #54, opened 2024-02-13
Added "voeg nieuw mandaat toe" form when editing mandataris

ruben-bottu:
image


Rahien:
I see what you're trying to do here. In the legacy view, you also can't pick the person when creating a new mandate for the person. However, I think we should still show it and even allow editing the person. I don't think it matters if they realize too late that they want to change the person when adding a new mandate?

This way, they will be reminded who they are giving the mandate to and it might even prevent mistakes?

Rahien:
Ok so when running this i see that the edit view appears inline, and you have to scroll down to see it. That makes it weird to show the person there indeed :thinking: 
What do you think about putting this form in a modal? Same for the editing of the end date then?

Rahien (approved):

PR #56, opened 2024-02-14
LMB-112 Person form

bfidlers:
Screenshot 2024-02-14 at 15 36 43
Example of person form within person-selector component, basic error validation is added, no required fields etc.


app/components/rdf-input-fields/geboorte-datum-input.js
```diff
@@ -0,0 +1,108 @@
...
+    super(...arguments);
+
+    const now = new Date();
+    const hundredYearsAgo = new Date(
+      now.getFullYear() - 100,
+      now.getMonth(),
+      now.getDate()
+    );
+    const eighteenYearsAgo = new Date(
+      now.getFullYear() - 18,
```
ruben-bottu, line 30: Maybe specify that 18 is the age of majority in Belgium?

Rahien: Keep this for now but i wonder if this is a correct requirement. Will check with Tim. (what happens if we are dealing with a jeugdraad or something?)

app/components/rdf-input-fields/geboorte-datum-input.js
```diff
@@ -0,0 +1,108 @@
...
+    const geboorte = await this.loadOrCreateGeboorte(isoDate, date);
+
+    replaceSingleFormValue(this.storeOptions, new NamedNode(geboorte.uri));
+
+    this.hasBeenFocused = true;
+  }
+
+  async loadOrCreateGeboorte(isoDate, date) {
+    let geboorte;
+    const queryResult = await this.store.query('geboorte', {
```
Rahien, line 73: this seems dangerous. It means that we reuse geboortes between person instances, which I wouldn't do.
Imagine someone wants to add additional info to the geboorte instance (like place of birth)

Suddenly you can't anymore.

I believe you will end up creating a new geboorte instance for every update (like we do for addresses).

We could get around this by making forms more complex, adding a nested object for geboorte (and address), but don't do this yet, it will take this story too far. I have another ticket for this.

app/components/rdf-input-fields/geboorte-datum-input.js
bfidlers: I think validation is something that should reviewed for all forms, so at that stage these requirements can also be revisited, and this works as a proof of concept for how to add custom validations for now.

app/components/rdf-input-fields/rijksregister-input.js
```diff
@@ -0,0 +1,90 @@
...
+    );
+
+    replaceSingleFormValue(this.storeOptions, new NamedNode(identificator.uri));
+
+    this.hasBeenFocused = true;
+  }
+
+  async loadOrCreateRijksregister(rrn) {
+    let identificator;
+    const queryResult = await this.store.query('identificator', {
```
Rahien, line 58: so in this case it's probably fine to query like this.

app/components/rdf-input-fields/rijksregister-input.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  updateValidations() {
+    super.updateValidations();
+
+    if (!this.rijksregisternummer) {
+      return;
+    }
+
+    // TODO ideally this should also check if the other fields are compatible with the rrn
```
Rahien, line 81: agreed not to do this right now

Rahien (changes requested):

app/components/rdf-input-fields/geboorte-datum-input.js
bfidlers: I'm still on the fence why we need to use a separate geboorte model, really curious what the thought was behind this. What you suggest would be a valid use case for implementing geboorte as a separate instance, still a meager one however. 

I saw it more like you want to group persons together based on their birthdate (for some unknown reason), but that indeed does not have much use. However this is also how it was done in legacy. 

But I can certainly change this If you want.

Rahien (approved):

PR #59, opened 2024-02-15
remove unused new person page

bfidlers:
Page was unused because of the new person-selector component.


ruben-bottu:
PR looks good, only woodpecker npm error

ruben-bottu (approved):

PR #61, opened 2024-02-19
Add Getters to Bestuursorgaan and Mandaat to Check If Decretaal

ruben-bottu:


app/models/bestuursorgaan.js
```diff
@@ -38,6 +38,28 @@ export default class BestuursorgaanModel extends Model {
...
   bevat;
 
+  async classificatieUri() {
+    const bestuursorgaan = await this.isTijdsspecialisatieVan;
+    return bestuursorgaan
+      ? await bestuursorgaan.classificatie.uri
+      : await this.classificatie.uri;
+  }
+
+  // TODO decretale bestuursorganen should be fetched from database
```
bfidlers, line 48: You also have a PR for the decretale concept schemes, so why isn't this fetched from the database?

app/models/bestuursorgaan.js
```diff
@@ -38,6 +38,28 @@ export default class BestuursorgaanModel extends Model {
...
+  // Bestuursorgaan and bestuursorgaan in de tijd are not the same. Use mandatendatabank schema for reference.
+  // go via isTijdsspecialisatieVan to go to the real bestuursorgaan.
```
bfidlers, line 50: seems to be a personal comment, not something that should be in the repo

app/models/mandaat.js
```diff
@@ -17,4 +17,18 @@ export default class MandaatModel extends Post {
...
+    const bestuursorgaanPositions = await this.bevatIn;
+    const areDecretaal = await Promise.all(
+      bestuursorgaanPositions.map((position) => position.isDecretaal())
+    );
+
+    return areDecretaal.some(Boolean);
+  }
+
+  // If any of the bestuursorganen is decretaal, then the mandaat is decretaal.
+  get isDecretaal() {
```
bfidlers, line 31: can a getter not be async? if it can, why is this extra function necessary?

app/models/bestuursorgaan.js
Rahien: i think it could be in this file, but all the way at the top as a general warning. This causes some major confusion when working with this type

app/models/mandaat.js
Rahien: it can be. however, when using this in templates, it will be a pain and will require promise helpers or the use of https://api.emberjs.com/ember/release/classes/PromiseProxyMixin. We should consider adding ember-resources here https://github.com/NullVoxPopuli/ember-resources/blob/main/docs/docs/resources.md

Rahien: will add a ticket to examine ember resources, ignore for now and use ember-promise-helpers when accessing this in templates

Rahien (approved):

PR #62, opened 2024-02-19
Add bestuursorgaan overview page

bfidlers:
This PR is mainly routing and some actual components. 
Is still missing some functionality and clean-up, but it should be possible to plug in functionality from other PR's in this base.

### DIFF
--- a/app/components/organen/organen-table.hbs
+++ b/app/components/organen/organen-table.hbs
@@ -1,0 +1,39 @@
+<AuDataTable
+  @content={{@content}}
+  @isLoading={{@isLoading}}
+  @noDataMessage="Geen bestuursorganen gevonden"
+  @sort={{@sort}}
+  @page={{@page}}
+  @size={{@size}}
+  as |t|
+>
+  <t.content as |c|>
+    <c.header>
+      <AuDataTableThSortable
+        @field="naam"
+        @currentSorting={{@sort}}
+        @label="Naam"
+      />
+      <th>Huidige voorzitter</th>
+      <th>Aantal leden</th>
+      <th>Oprichtingsdatum</th>
+      <th>{{! Bewerk Action }}</th>
+      <th>{{! Archive Action }}</th>
+    </c.header>
+
+    <c.body as |row|>
+      <td>{{row.naam}}</td>
+      <td>TODO add voorzitter</td>
+      <td>TODO add aantal leden</td>
+      <td>
+        {{moment-format row.bindingStart "DD-MM-YYYY"}}</td>
+      <td>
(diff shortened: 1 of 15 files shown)


app/components/organen/organen-table.hbs
```diff
@@ -0,0 +1,39 @@
...
+      <th>{{! Bewerk Action }}</th>
+      <th>{{! Archive Action }}</th>
+    </c.header>
+
+    <c.body as |row|>
+      <td>{{row.naam}}</td>
+      <td>TODO add voorzitter</td>
+      <td>TODO add aantal leden</td>
+      <td>
+        {{moment-format row.bindingStart "DD-MM-YYYY"}}</td>
```
Rahien, line 29: oprichtingsdatum isn't the same as the bindingStart. Binding Start is the start of the current legislature (bestuursorgaan in de tijd). I would remove this field for now as we don't have it in the model either

app/routes/organen.js
```diff
@@ -0,0 +1,36 @@
...
+    //   this.router.transitionTo('index');
+  }
+
+  async model() {
+    const bestuurseenheid = this.currentSession.group;
+    const bestuursorganen = await this.getAllBestuursorganen(
+      bestuurseenheid.get('id')
+    );
+
+    return RSVP.hash({
```
Rahien, line 25: not paginating here. i think that is fine and that we don't expect too many bestuursorganen

app/routes/organen.js
```diff
@@ -0,0 +1,36 @@
...
+    this.session.requireAuthentication(transition, 'login');
+
+    // TODO we probably want some access rights here.
+    // if (!this.currentSession.canAccessBestuursorganen)
+    //   this.router.transitionTo('index');
+  }
+
+  async model() {
+    const bestuurseenheid = this.currentSession.group;
+    const bestuursorganen = await this.getAllBestuursorganen(
```
Rahien, line 21: i think we should filter here and only get the bestuursorganen that aren't a 'tijdspecialisatieVan' of something else (i.e. not the bestuursorganen in de tijd)

Rahien: fine to do this in a separate pr though! this structure is great to have already!

Rahien (approved):

app/components/organen/organen-table.hbs
bfidlers: Would it be a good idea to check the bestuursorganen in de tijd for the earliest start? or not that important now?

app/routes/organen.js
bfidlers: this also needs some extra filtering with regard to activeness, will do this in the next PR

PR #63, opened 2024-02-19
Karel/lmb 115 create bestuursorgaan

Rahien:
image


ruben-bottu:
Reminder that this branch should also be checked out in the backend

app/controllers/organen/beheer/new.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  @action
+  cancel() {
+    this.router.transitionTo('mandatenbeheer.bestuursorganen');
+  }
+
+  @action
+  onCreate({ instanceId }) {
+    this.router.transitionTo('mandatenbeheer.bestuursorganen.edit', instanceId);
```
ruben-bottu, line 22: These need to transition to the new routes, error: `The route mandatenbeheer.bestuursorganen.edit was not found`

Rahien: well spotted!

ruben-bottu (approved):

PR #64, opened 2024-02-20
LMB-120 Bestuursorgaan page

bfidlers:
This should cover everything from LMB-120


app/components/organen/inactive-orgaan-table.hbs
```diff
@@ -0,0 +1,31 @@
...
+  <t.content as |c|>
+    <c.header>
+      <AuDataTableThSortable
+        @field="naam"
+        @currentSorting={{@sort}}
+        @label="Naam"
+      />
+      <th>Aantal leden</th>
+      <th>Oprichtingsdatum</th>
+      <th>Einddatum</th>
```
Rahien, line 17: so this is the reason for having different tables right? the columns are not the same?

app/controllers/organen/beheer/index.js
```diff
@@ -1,18 +1,101 @@
...
+  active_sort = 'naam';
+  inactive_sort = 'naam';
 
   @action
   createNewOrgaan() {
     this.router.transitionTo('organen.beheer.new');
   }
+
+  @action
+  async getNbMembers(orgaan) {
```
Rahien, line 19: could this be a method on the bestuursorgaan model?

app/components/organen/inactive-orgaan-table.hbs
bfidlers: yes indeed
it might be possible to work with an if clause, or with some general component which both tables extend, but I think this is the best way for now, as we don't now yet how these tables might change.
It might be a good idea to make a story for a future attempt to generalise all tables.

app/controllers/organen/beheer/index.js
bfidlers: Not sure if that is possible, might be difficult with the async, but I could try if you want.

app/routes/organen/beheer/index.js
```diff
@@ -5,42 +5,86 @@ import { action } from '@ember/object';
...
-    const bestuursorganen = await this.store.query('bestuursorgaan', options);
+    const active_options = this.getActiveOptions(
+      params,
+      parentModel.bestuurseenheid
+    );
+    const activeOrganenUnfiltered = await this.store.query(
+      'bestuursorgaan',
+      active_options
+    );
+    const activeOrganen = [];
```
Rahien, line 25: the model function is getting long. i would extract line 25 till 33 to a function called getActiveDecretaleOrganen or something

PR #65, opened 2024-02-20
Added Styling for Readonly View of Forms

ruben-bottu:
image

### DIFF
--- a/app/components/form/instance.hbs
+++ b/app/components/form/instance.hbs
@@ -15,7 +15,7 @@
             @forceShowErrors={{true}}
             @useNewListingLayout={{true}}
             @level={{2}}
-            class="au-u-max-width-medium"
+            class="au-u-max-width-medium {{if this.isEditable '' 'disabled'}}"
           />
           {{yield}}
         </section>
--- a/app/components/form/instance.hbs
+++ b/app/components/form/instance.hbs
@@ -15,7 +15,7 @@
             @forceShowErrors={{true}}
             @useNewListingLayout={{true}}
             @level={{2}}
-            class="au-u-max-width-medium {{if this.isEditable '' 'disabled'}}"
+            class="au-u-max-width-medium {{unless this.isEditable 'disabled'}}"
           />
           {{yield}}
         </section>
--- a/app/styles/_shame.scss
+++ b/app/styles/_shame.scss
@@ -148,15 +148,8 @@
   display: none;
 }
 
-.disabled
-  .is-disabled
-  .au-c-toggle-switch__input:checked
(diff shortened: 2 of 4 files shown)


app/components/form/instance.hbs
```diff
@@ -15,7 +15,7 @@
             @forceShowErrors={{true}}
             @useNewListingLayout={{true}}
             @level={{2}}
-            class="au-u-max-width-medium"
+            class="au-u-max-width-medium {{if this.isEditable '' 'disabled'}}"
```
bfidlers, line 18: I find this ternary if here a bit confusing and it doesn't seem necessary, I would suggest using 
`{{unless this.isEditable 'disabled'}}`

app/templates/organen/orgaan.hbs
```diff
@@ -20,15 +20,20 @@
...
+  />
+
+  <div class="au-o-layout">
+    <AuButton
+      class="au-u-margin-bottom"
+      @skin="secondary"
+      {{on "click" this.edit}}
+    >Pas aan</AuButton>
 
-<AuToolbar @border="top" @size="large" as |Group|>
```
bfidlers, line 29: Why did you remove the toolbar here, seems like a good practice to use those?

app/styles/_shame.scss
bfidlers, line 1: We will probably need to review these rules when we use disabled forms with other fields, but looks good for now.

bfidlers (approved):
Few comments, but otherwise looks okay.

app/components/form/instance.hbs
ruben-bottu: Oh didn't know you could do that, thanks!

app/templates/organen/orgaan.hbs
ruben-bottu: Because we wanted to make the "Pas aan" look inline with the form. It might be strange to have it in the toolbar as well no?

ruben-bottu: And karel wanted the todo to be directly under the form

app/styles/_shame.scss
```diff
@@ -127,3 +127,31 @@
...
+  background-color: transparent;
+  color: black;
+  border: transparent;
+  padding-left: 0;
+}
+
+.disabled .ember-power-select-trigger[aria-disabled="true"] {
+  background-color: transparent;
+  border: transparent;
+  padding-left: 0 !important;
```
ruben-bottu, line 144: This !important is sadly necessary, because it is also !important in ember-submission-form-fields

bfidlers (approved):

PR #66, opened 2024-02-21
Moved Container Styling Out Of instance and new-instance Components

ruben-bottu:
Moved au-c-main-container and other padding and margin styling to outside the instance and new-instance component.
image


Rahien (approved):
awesome!

PR #67, opened 2024-02-21
LMB-169 mandataris table

bfidlers:
Cleaned up and now with search and everything.

See below an example of the page with table, pagination, search and bestuursperiodenselector.
Screenshot 2024-02-21 at 15 51 15

### DIFF
--- a/app/controllers/organen/orgaan/mandatarissen.js
+++ b/app/controllers/organen/orgaan/mandatarissen.js
@@ -27,9 +27,8 @@ export default class OrganenMandatarissenController extends Controller {
     const queryParams = {
       page: 0,
       startDate: period.startDate,
+      endDate: period.endDate,
     };
-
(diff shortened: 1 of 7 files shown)


app/controllers/organen/orgaan/mandatarissen.js
```diff
@@ -0,0 +1,37 @@
...
+
+  @restartableTask
+  *search(searchData) {
+    yield timeout(SEARCH_TIMEOUT);
+    this.page = 0;
+    this.filter = searchData;
+  }
+
+  @action
+  selectPeriod(period) {
```
Rahien, line 26: for now this is enough, but i think this is supposed to be in the orgaan route too. That complicates things because you need to transition to the correct new (sub) route based on the one you are in right now. Not sure how easy -> new story

Rahien (approved):

PR #72, opened 2024-02-23
allow edit of mandaten

Rahien:
image

note: it is currently not possible to set the value of a number input back to undefined. We do want that I believe... Created another story for this


app/templates/mandaat/edit.hbs
```diff
@@ -1,3 +1,24 @@
...
+    <Form::Instance
+      @instanceId={{this.model.instanceId}}
+      @form={{this.model.form}}
+      @onSave={{this.onSave}}
+    />
+  </div>
+</div>
 {{outlet}}
+
+<AuToolbar @border="top" @size="large" as |Group|>
```
bfidlers, line 20: Might be nice to just give this onCancel action to the form instance, than you have the two buttons next to each other. But depends on preference of course.

bfidlers (approved):
One small comment, but is more a personal preference.
Looks good otherwise.

app/templates/mandaat/edit.hbs
Rahien: fair point!

PR #74, opened 2024-02-23
Allow Creating Bestuursorganen In De Tijd (niet decretaal)

ruben-bottu:
image


app/controllers/organen/orgaan/mandatarissen.js
```diff
@@ -34,4 +34,9 @@ export default class OrganenMandatarissenController extends Controller {
       queryParams,
     });
   }
+
+  @action
+  handleAddBestuursperiodeClick() {
+    this.router.transitionTo('organen.orgaan.bestuursperiodes.new');
```
Rahien, line 40: this isn't the case anymore right?

app/templates/organen/orgaan/mandatarissen.hbs
```diff
@@ -5,6 +5,9 @@
       @selectedPeriod={{@model.selectedPeriod}}
       @onSelect={{this.selectPeriod}}
     />
+    {{#unless (await this.model.bestuursOrgaan.isDecretaal)}}
+      <BestuursperiodeNewModal @bestuursorgaan={{this.model.bestuursOrgaan}} />
```
Rahien, line 9: this looks like it always shows the modal?

app/controllers/organen/orgaan/mandatarissen.js
Rahien: the action also doesn't get called

app/templates/organen/orgaan/mandatarissen.hbs
Rahien: I get it! Your component is actually the BestuursperiodeNewButton, that opens a modal on click. Could you rename as such?

Rahien (changes requested):

PR #81, opened 2024-02-26
Ben/lmb 176 person page

bfidlers:
New person page, with to do components that will transition to different pages.
This page is linked in the mandataris search page and in the mandataris table in the orgaan page.

Screenshot 2024-02-26 at 16 18 05

Old page is still available, because some logic will need to be reintroduced, but this is in another story, if LMB-191 is done, this shall disappear.


app/templates/mandatarissen/persoon.hbs
```diff
@@ -0,0 +1,64 @@
...
+  <div class="au-o-box">
+    <AuHeading @level="3" @skin="4">
+      Huidige mandaten
+    </AuHeading>
+
+    <AuHr />
+
+    <ul class="au-c-list-vertical">
+      {{#each @model.mandatarissen as |mandataris|}}
+        {{#if mandataris.isActive}}
```
Rahien, line 23: could you reference this code in the story about adding fallbacks to these lists in the old view? LMB-189

Rahien (approved):

PR #85, opened 2024-02-27
Layout bestuursorgaan pagina

ruben-bottu:
image

### DIFF
--- a/app/components/mandatenbeheer/bestuursorgaan-mandaten.hbs
+++ b/app/components/mandatenbeheer/bestuursorgaan-mandaten.hbs
@@ -1,4 +1,7 @@
-<div class="au-o-box no-pagination" {{did-insert this.initialize.perform}}>
+<div
+  class="au-o-box no-pagination"
+  {{did-update this.initialize.perform @mandaten}}
+>
   <AuDataTable
     @content={{this.orderedMandaten}}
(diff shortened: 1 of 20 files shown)


ruben-bottu:
Sadly mandaten overview still breaks if you switch to it from details and mandatarissen tabs

app/components/mandatenbeheer/bestuursorgaan-mandaten.hbs
```diff
@@ -1,4 +1,7 @@
-<div class="au-o-box no-pagination" {{did-insert this.initialize.perform}}>
+<div
+  class="au-o-box no-pagination"
+  {{did-update this.initialize.perform @mandaten}}
```
Rahien, line 3:
```suggestion
  {{did-update this.initialize.perform @mandaten}}
  {{did-insert this.initialize.perform}}
```

Rahien: you need both or on first render the orderedMandaten is not computed

Rahien: like we said, it would be much more clean to do this in the route but i think you created a story for that?

Rahien (changes requested):

Rahien (approved):

PR #86, opened 2024-02-28
Ben/lmb 192 mandataris page

bfidlers:
Screenshot 2024-02-28 at 00 13 51
Screenshot 2024-02-28 at 00 14 16


Rahien (approved):
looks good! i think we should rework the flow from the mandataris table(s). It says bewerk right now but it brings you to the person detail page. Either we should change the CTA or bring them to the mandaat details.
Still, this is another story

PR #95, opened 2024-03-01
add undefined check fractie

bfidlers:


app/utils/form-business-rules/mandataris-membership.js
```diff
@@ -62,6 +62,9 @@ export const syncMandatarisMembership = async function (
 ) {
   const { begin: begin, end } = getMandatarisDates(mandatarisUri, storeOptions);
   const membership = await fetchMembership(store, mandatarisUri, storeOptions);
+  if (!membership) {
+    return;
+  }
```
ruben-bottu, line 67: My question is: do we want to be able to have a mandataris without a fractie? We have Onafhankelijk for those who are not connected to a political party

ruben-bottu (approved):

PR #96, opened 2024-03-01
add extra checks if start or einde mandataris undefined

bfidlers:


ruben-bottu:
image
The error message that is shown is still quite confusing, but at least it doen't crash anymore

ruben-bottu (approved):

PR #98, opened 2024-03-01
Ben/fix deprecations

bfidlers:
Fix some deprecations in the person search form.

### DIFF
--- a/app/components/person/search-form.js
+++ b/app/components/person/search-form.js
@@ -42,6 +42,28 @@ export default class SharedPersoonPersoonSearchFormComponent extends Component {
   }
 
   @restartableTask
+  *setVoornaam(searchData) {
+    yield timeout(300);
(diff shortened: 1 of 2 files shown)


app/components/person/search-form.js
```diff
@@ -41,6 +41,30 @@ export default class SharedPersoonPersoonSearchFormComponent extends Component {
...
+  @action
+  setAchternaam(event) {
+    if (event && typeof event.preventDefault === 'function')
+      event.preventDefault();
+    this.achternaam = event.target.value.trim();
+    this.search.perform();
+  }
+
+  @action
+  setRrn(event) {
```
ruben-bottu, line 61: I would still call the method `setRijksregisternummer`. What Rrn means can be derived from context, but usually names in full hold up better for readability.

ruben-bottu (approved):

PR #105, opened 2024-03-05
improve update mandataris state

Rahien:
image

image

image

image


app/components/mandatarissen/fractie-selector.js
```diff
@@ -19,27 +21,34 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
...
         },
       },
     };
+    if (searchData) {
+      queryParams.filter.naam = searchData;
+    }
     let fracties = yield this.store.query('fractie', queryParams);
     fracties = fracties.filter((f) => !f.get('fractietype.isOnafhankelijk'));
     //sets dummy
-    if ('onafhankelijk'.includes(searchData.toLowerCase())) {
```
bfidlers, line 40: This whole isOnafhankelijk logic seems broken. If I understand it correctly a new fractie should be created if the person is onafhankelijk. But this is not the case at all, this only happens if you search for onafhankelijk. Which seems to give more than one fractie btw. And even then this fractie is never saved.

app/components/mandatarissen/update-state.hbs
```diff
@@ -57,6 +57,46 @@
...
+  {{#unless this.isTerminatingMandate}}
+    <div>
+      <AuLabel>
+        Beleidsdomeinen
+      </AuLabel>
+      <Mandatarissen::BeleidsdomeinSelectorWithCreate
+        @beleidsdomeinen={{this.selectedBeleidsdomeinen}}
+        @onSelect={{this.updateBeleidsdomeinen}}
+      />
+      <AuHelpText>Enkel voor burgemeester en schepenen</AuHelpText>
```
bfidlers, line 69: If this is true, this is really something we should clarify in other places as well, or limit with form restrictions or something like that.

app/components/mandatarissen/update-state.hbs
```diff
@@ -57,6 +57,46 @@
...
+    </div>
+    <div>
+      <AuLabel>
+        Fractie
+      </AuLabel>
+      {{#if this.loading}}
+        Loading...
```
bfidlers, line 76: Bit weird that this has a separate loading message, since the beleidsdomeinselectorwithcreate does not and since this component contains a loading state.

app/components/mandatarissen/update-state.js
```diff
@@ -7,23 +7,39 @@ import { showErrorToast, showSuccessToast } from 'frontend-lmb/utils/toasts';
...
+    return this.load.isRunning;
+  }
+
+  @dropTask
+  *load() {
+    this.newStatus = this.args.mandataris.status;
+    this.date = new Date();
+    this.selectedBeleidsdomeinen = this.args.mandataris.beleidsdomein.slice();
+    this.rangorde = this.args.mandataris.rangorde;
+    this.selectedFractie = yield (yield this.args.mandataris.heeftLidmaatschap)
```
bfidlers, line 35: painful, isn't this solvable as well with the get('..') instead? But then again, wouldn't really make it much better.

bfidlers (approved):
It doesn't really make sense that the "wijzig huidige situatie" and "corrigeer fouten" have different components for the beleidsdomeinen-selector. This feels really broken (also the create mandataris form uses the component that can only select one).

Few remarks here and there, otherwise looks good.

app/components/mandatarissen/beleidsdomein-selector-with-create.js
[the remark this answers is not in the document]
Rahien: the todo requires moving to ember data v5 which is something non-trivial and i don't want to attempt for now. the other thing deserves a new story

app/components/mandatarissen/update-state.hbs
Rahien: i agree that we should limit this, i'm doing a trial for things like this in my final story in this 'saga', but didn't want to add it to existing things yet

app/components/mandatarissen/update-state.js
Rahien: indeed

app/components/legacy/mandatenbeheer/fractie-selector.hbs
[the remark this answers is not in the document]
bfidlers: Yeah sure, fine for me.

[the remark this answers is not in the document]
bfidlers: Or let's discuss this at the sprint planning maybe?

app/components/mandatarissen/update-state.hbs
Rahien: the reason is that this is the only thing that really can't be used while the component is loading, all other fields can be filled out

PR #106, opened 2024-03-06
add pagination to form instances

Rahien:
image

### DIFF
--- a/app/components/form/instance-table.js
+++ b/app/components/form/instance-table.js
@@ -41,16 +41,45 @@ export default class InstanceTableComponent extends Component {
     }
   }
 
+  @action
   async onInit() {
     const form = this.args.formDefinition;
     const id = form.id;
-    const response = await fetch(`/form-content/${id}/instances`);
+    const response = await fetch(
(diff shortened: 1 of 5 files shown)


app/components/form/instance-table.js
```diff
@@ -41,16 +41,45 @@ export default class InstanceTableComponent extends Component {
     }
   }
 
+  @action
   async onInit() {
```
bfidlers, line 45: This seems more complicated than it should be, I think you could rely more on ember to track the whole pagination etc. but I don't really have time to investigate it now.

bfidlers: or maybe not since we don't use resources, I'll have a look at that on Friday.

Rahien: sadly no, the data table assumes the same meta extensions exist like they were done for ember data relations/queries

bfidlers (approved):
Bit sad about the manual pagination, but unfortunately that is a result of not being able to use ember resources.

Looks good otherwise.

PR #111, opened 2024-03-08
workaround for resource cache invalidation

Rahien:


app/routes/mandatarissen/mandataris.js
```diff
@@ -36,6 +36,9 @@ export default class MandatarissenMandatarisRoute extends Route {
         'bekleedt.bestuursfunctie',
         'bekleedt.bevat-in',
         'beleidsdomein',
+        'status',
```
Rahien, line 39: this ensures that when we refresh the model, these relations are also refreshed

app/utils/toasts.js
```diff
@@ -2,6 +2,7 @@ const notifyFormSavedSuccessfully = (toaster) => {
   toaster.notify('Het formulier werd correct opgeslagen.', 'Succes', {
     type: 'success',
     icon: 'circle-check',
+    timeOut: 5000,
```
Rahien, line 5: got tired of seeing the success messages forever heaping up in the top right corner

app/components/mandatarissen/mandataris-edit-prompt.js
```diff
@@ -59,7 +60,7 @@ export default class MandatenbeheerMandatarisEditPromptComponent extends Compone
       sourceGraph: SOURCE_GRAPH,
     });
 
-    setTimeout(() => this.args.mandataris.reload(), 1000);
+    setTimeout(() => this.router.refresh(), 1000);
```
Rahien, line 63: refresh reloads all models up to this point, which is safer. It also means that if we request relations to be included those will be reloaded too

bfidlers (approved):
Looks good

PR #115, opened 2024-03-12
bump appuniversum to 3.2.0

Rahien:


bfidlers (approved):
Unfortunate you had to refactor legacy components, looking forward to next week when we remove this.
Unfortunately we now also have deprecation warnings about app-universum 4.0, but it looks contained.

Looked good otherwise.

PR #117, opened 2024-03-12
fix modal over modal

Rahien:
bump appuniversum further to have modals over modals, make exit confirm modal smaller

image


bfidlers (approved):
Looks good

PR #118, opened 2024-03-12
Added form and form-extension models

ruben-bottu:


app/models/form-extension.js
```diff
@@ -0,0 +1,11 @@
+import Model, { belongsTo } from '@ember-data/model';
+
+export default class FormExtensionModel extends Model {
+  @belongsTo('form', {
+    async: true,
+    inverse: 'form-extension',
+    polymorphic: true,
+    as: 'form-extension',
+  })
+  classificatie;
```
Rahien, line 10: this seems wrong?

app/models/form-extension.js
```diff
@@ -0,0 +1,11 @@
+import Model, { belongsTo } from '@ember-data/model';
+
+export default class FormExtensionModel extends Model {
+  @belongsTo('form', {
+    async: true,
+    inverse: 'form-extension',
+    polymorphic: true,
+    as: 'form-extension',
```
Rahien, line 8: i don't think you need this

app/models/form.js
```diff
@@ -0,0 +1,18 @@
...
+  @attr targetLabel;
+  @attr prefix;
+  @attr id;
+  @attr formTtl;
+  @attr metaTtl;
+
+  @hasMany('form-extension', {
+    async: true,
+    inverse: 'form',
+    polymorphic: true,
```
Rahien, line 14: this isn't polymorphic, it can only be form extensions

Rahien (changes requested):

Rahien (approved):

PR #122, opened 2024-03-14
Add Election Results Models

ruben-bottu:


app/models/bestuursorgaan.js
```diff
@@ -41,6 +41,12 @@ export default class BestuursorgaanModel extends Model {
   })
   isTijdsspecialisatieVan;
 
+  @belongsTo('rechtstreekse-verkiezing', {
```
Rahien, line 44: same as in the app pr: would rename to simply verkiezing

app/models/lijsttype.js
```diff
@@ -0,0 +1,13 @@
+import Model, { attr } from '@ember-data/model';
+
+export default class LijsttypeModel extends Model {
+  @attr uri;
+  @attr label;
+  @attr scopeNote;
```
Rahien, line 6: same as in app pr: would remove

app/models/verkiezingsresultaat-gevolg-code.js
```diff
@@ -0,0 +1,13 @@
+import Model, { attr } from '@ember-data/model';
+
+export default class VerkiezingsresultaatGevolgCodeModel extends Model {
+  @attr uri;
+  @attr label;
+  @attr scopeNote;
```
Rahien, line 6: same as in app pr: would remove

Rahien (approved):
just some things to bring inline with the app prs

PR #123, opened 2024-03-15
LMB-244 Add check dirty state mandataris

bfidlers:
…s been selected

### DIFF
--- a/app/components/mandatarissen/update-state.js
+++ b/app/components/mandatarissen/update-state.js
@@ -14,6 +14,7 @@ export default class MandatarissenUpdateState extends Component {
   @tracked bestuursorganenForFractie = [];
   @tracked rangorde = null;
   @tracked selectedReplacement = null;
+  @tracked replacementSelected;
 
   @service mandatarisStatus;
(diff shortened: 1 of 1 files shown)


app/components/mandatarissen/update-state.js
```diff
@@ -14,6 +14,7 @@ export default class MandatarissenUpdateState extends Component {
   @tracked bestuursorganenForFractie = [];
   @tracked rangorde = null;
   @tracked selectedReplacement = null;
+  @tracked replacementSelected;
```
Rahien, line 17: Fine for now but if we are going to align everything on multiple replacements, we should change this from a boolean to a list of ids so that we can also track if there are no changes

Rahien (approved):

app/components/mandatarissen/update-state.js
bfidlers: The variable name is maybe a bit confusing, the idea was more if a change happens, set this variable to true, but now with this component you can only select a replacement. Will update the name.

PR #124, opened 2024-03-19
form history

Rahien:
image

To be able to test this pr, you will need this branch in the form-content service: https://github.com/lblod/form-content-service/pull/34

note: when reverting the mandataris-edit form, an exception is thrown in the console, coming from the form library. Some hook that isn't cleaned up properly. The application keeps working as expected though, so i will just add a bug ticket to our backlog and ignore for now.

### DIFF
--- a/app/components/form/history.hbs
+++ b/app/components/form/history.hbs
@@ -1,0 +1,35 @@
+{{#if this.loading}}
+  Loading...
+{{else}}
+  <AuTable @size={{this.size}}>
+    <:header>
+      <tr>
+        <th>Tijdstip</th>
+        <th>Gebruiker</th>
+        <th>Aanpassing</th>
+        <th></th>
+      </tr>
+    </:header>
+    <:body>
+      {{#each this.history as |historyItem|}}
+        <Form::HistoryRow
+          @historyItem={{historyItem}}
+          @onRestore={{perform this.restoreHistoryItem}}
+        />
+      {{else}}
+        <tr>
+          <td>Geen aanpassingen gevonden</td><td></td><td></td><td></td>
+        </tr>
+      {{/each}}
+    </:body>
+  </AuTable>
+  <Shared::TablePagination
+    @pageSize={{this.size}}
+    @page={{this.page}}
+    @totalItems={{this.totalCount}}
+    @toNextPage={{this.toNextPage}}
+    @toPreviousPage={{this.toPreviousPage}}
+  />
+
+  {{yield}}
+{{/if}}
--- a/app/components/form/history.hbs
+++ b/app/components/form/history.hbs
@@ -18,18 +18,20 @@
         />
       {{else}}
         <tr>
-          <td>Geen aanpassingen gevonden</td><td></td><td></td><td></td>
+          <td colspan="4">Geen aanpassingen gevonden</td>
         </tr>
       {{/each}}
     </:body>
   </AuTable>
-  <Shared::TablePagination
-    @pageSize={{this.size}}
-    @page={{this.page}}
-    @totalItems={{this.totalCount}}
-    @toNextPage={{this.toNextPage}}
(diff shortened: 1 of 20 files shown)


app/components/form/history.js
```diff
@@ -0,0 +1,71 @@
...
+
+    const userIds = new Set([...history.map((h) => h.creator)]);
+    const users = yield this.store.query('gebruiker', {
+      filter: {
+        id: Array.from(userIds).join(','),
+      },
+    });
+
+    this.history = history.map((h) => {
+      const user = users.findBy('id', h.creator);
```
bfidlers, line 40: This findBy throws a deprecation.

app/components/form/instance.js
```diff
@@ -111,9 +112,16 @@ export default class InstanceComponent extends Component {
     this.args.onCancel();
   }
 
-  async onInit() {
+  @action
+  async onRestore(historicalInstance) {
```
bfidlers, line 116: Shouldn't restoring persist? Which it does not appear to do know.

bfidlers (commented):
Looks good and some really nice additions. A few remarks however:

1. I get a few bugs, a first one is if you try to load a form with formHistory enabeled that does not have any changes yet:
Screenshot 2024-03-19 at 10 55 43
Another one just happens in the mandataris-edit, I think this is the same as what you mentioned in the description?
Screenshot 2024-03-19 at 10 55 20
2. The UI is a bit confusing, it does allow to restore the change you just made, which actually does nothing? But you can't revert this change (e.g. if you make your first change you can't go back to the state before this change)? Also there should be a way to preview what the state was, but this should be a different story.
3. It is possible to just keep on updating a form without actually changing things, this should be checked, probably in the service, but this could also be a different story.

app/components/form/instance.js
Rahien: so indeed it doesn't but this was intentional as we don't have a preview/show changes feature yet. So the 'terugzetten' button currently only sets the values in the form but doesn't save. I have now tried to make this more obvious by only enabling the update button when the form has changes (which it has if you restore a previous version)

Also, you may have had the feeling that hitting the button didn't do anything because the current version is the top one, so you'd reset it to itself. i have removed the button in that case and marked is as the current version. 

Legacy data (not created by us) does not have an initial version (didn't go through the form content service. I will create another story to have a fallback for such cases, where the form content service creates a history item before updates if the current item doesn't have a history yet.

Rahien:
> Looks good and some really nice additions. A few remarks however:
> 
> 1. I get a few bugs, a first one is if you try to load a form with formHistory enabeled that does not have any changes yet:
>    Screenshot 2024-03-19 at 10 55 43
>    Another one just happens in the mandataris-edit, I think this is the same as what you mentioned in the description?
>    Screenshot 2024-03-19 at 10 55 20
> 2. The UI is a bit confusing, it does allow to restore the change you just made, which actually does nothing? But you can't revert this change (e.g. if you make your first change you can't go back to the state before this change)? Also there should be a way to preview what the state was, but this should be a different story.
> 3. It is possible to just keep on updating a form without actually changing things, this should be checked, probably in the service, but this could also be a different story.

1. fixed (the second one is indeed the one i mentioned in the description)
2. and 3. should be fixed by the actions i put in the restoring comment

app/components/form/history.hbs
```diff
@@ -0,0 +1,38 @@
...
+        <th>Aanpassing</th>
+        <th></th>
+      </tr>
+    </:header>
+    <:body>
+      {{#each this.history as |historyItem index|}}
+        <Form::HistoryRow
+          @historyItem={{historyItem}}
+          @onRestore={{perform this.restoreHistoryItem}}
+          @isCurrentVersion={{eq index 0}}
```
bfidlers, line 18: This does not work if you have multiple pages
Screenshot 2024-03-19 at 16 30 41

PR #128, opened 2024-03-19
fix auinput component

bfidlers:
Fix auinput value bug

### DIFF
--- a/app/components/mandatarissen/update-state.js
+++ b/app/components/mandatarissen/update-state.js
@@ -245,6 +245,13 @@ export default class MandatarissenUpdateState extends Component {
     this.newStatus = status;
   }
 
+  @action
+  updateRangorde(event) {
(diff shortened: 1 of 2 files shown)


app/components/mandatarissen/update-state.js
```diff
@@ -245,6 +245,13 @@ export default class MandatarissenUpdateState extends Component {
     this.newStatus = status;
   }
 
+  @action
+  updateRangorde(event) {
+    if (event && typeof event.preventDefault === 'function')
```
Rahien, line 250: please always use {} in ifs

Rahien (approved):

app/components/mandatarissen/update-state.js
bfidlers: oeps, seems to be in a few places, I'll fix this in a different PR

PR #135, opened 2024-03-20
Show installatievergaderingen

ruben-bottu:
image


app/routes/verkiezingen.js
```diff
@@ -0,0 +1,102 @@
...
+      yearToStatusIdToIvs[year][statusId].push(vergadering);
+
+      statusIdToStatus[statusId] = status;
+    });
+
+    return { yearToStatusIdToIvs, statusIdToStatus };
+  }
+
+  /*
+  [
```
Rahien, line 45: add a comment explaining that this is the format

Rahien (approved):

app/routes/verkiezingen.js
ruben-bottu: Fixed in next PR

PR #137, opened 2024-03-20
group mandataris instances in mandataris page of bestuursorgaan

Rahien:
image

### DIFF
--- a/app/routes/organen/orgaan/mandatarissen.js
+++ b/app/routes/organen/orgaan/mandatarissen.js
@@ -79,7 +79,7 @@ export default class OrganenMandatarissenRoute extends Route {
         result = -1;
       }
 
-      if (params.sort.indexOf('-') >= 0) {
+      if (params.sort.indexOf('-') === 0) {
         return result * -1;
       }
       return result;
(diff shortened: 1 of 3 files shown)


app/routes/organen/orgaan/mandatarissen.js
```diff
@@ -23,10 +24,71 @@ export default class OrganenMandatarissenRoute extends Route {
...
+        const personId = (await mandataris.isBestuurlijkeAliasVan).id;
+        const mandaatId = (await mandataris.bekleedt).id;
+        const key = `${personId}-${mandaatId}`;
+        const existing = personMandaatData[key];
+        if (existing) {
+          existing.foldedStart = moment.min(
+            moment(existing.foldedStart),
+            moment(mandataris.start)
+          );
+          if (!mandataris.einde || !existing.foldedEnd) {
```
bfidlers, line 45: Not sure why this is here for end and not for start. Probably we want to require a start, but to be completely save, I would not presume this here.

bfidlers (commented):
Looks good and seems to work, but this breaks sorting of dates. Sorting of names still works, mandates as well, fracties kind of.

bfidlers (commented):
Seems good for dates, some weird shananigans going on with fractie however.

bfidlers (approved):

PR #139, opened 2024-03-22
Show election results

ruben-bottu:
image


app/router.js
bfidlers, line 1: I think verkiezingsuitslag should be nested under verkiezingen, since this is also the route you will be taking. If you want to seperate these pages and not actually nest the model routing etc, you can put what is now under verkiezingen in an index page.

app/routes/mandatarissen/search.js
```diff
@@ -28,12 +28,13 @@ export default class MandatarissenSearchRoute extends Route {
         number: params.page,
         size: params.size,
       },
-      filter: params.filter,
       'filter[:has:is-aangesteld-als]': true,
       include:
         'is-aangesteld-als,is-aangesteld-als.bekleedt,is-aangesteld-als.bekleedt.bestuursfunctie',
     };
-
+    if (params.filter && params.filter.length > 0) {
```
bfidlers, line 35: I really don't think this should be here. This has nothing to do with this PR.

bfidlers (commented):
I really think the routing should be different, these routes are in the same module, so the routing should also be grouped.

bfidlers:
Oh one more comment, it really isn't clear that the selector actually updates the status, this looks more like a search option than an actual action.

bfidlers (approved):

PR #140, opened 2024-03-22
Verkiezingen menus and breadcrumbs

ruben-bottu:
image


bfidlers (approved):
I would update the icon that is displayed, maybe vote?
But otherwise looks good.

PR #154, opened 2024-04-03
Remove geboorte datum and rijksregister input components

ruben-bottu:
Do an `npm install` when checking out this PR!


Rahien (approved):

app/components/rdf-input-fields/rijksregister-input.hbs
```diff
@@ -1,30 +0,0 @@
...
-</AuLabel>
-
-<AuInput
-  @error={{this.hasErrors}}
-  @disabled={{@show}}
-  @width={{if @inTable "block"}}
-  @warning={{this.hasWarnings}}
-  id={{this.inputId}}
-  value={{this.rijksregisternummer}}
-  {{au-inputmask
```
Rahien, line 17: i think we need a very basic component that still enforces this input mask

Rahien (changes requested):
we should have a component for the rijksregister input still (albeit much more basic than before)

Rahien (approved):

PR #156, opened 2024-04-05
Replace mandataris fractie selector

ruben-bottu:


app/components/rdf-input-fields/mandataris-fractie-selector.js
```diff
@@ -68,42 +69,37 @@ export default class MandatarisFractieSelector extends InputFieldComponent {
...
+    const uri = fractie.uri;
 
-    // Note: this process creates a new membership whenever the value of the fractie selector changes.
-    // For history purposes, we shouldn't delete the old membership so if we want to view a previous
-    // version of the mandataris forms, we can still see the old membership.
-    // Memberships should therefore also never be modified, only created and deleted.
-    replaceSingleFormValue(this.storeOptions, new NamedNode(newMembership.uri));
+    replaceSingleFormValue(this.storeOptions, uri ? new NamedNode(uri) : null);
+    this.selectedFractie = fractie;
+    // TODO Why is this useful? Also hasn't been declared as a field.
```
Rahien, line 92: this is from the inputfield component that we are extending. the forms use it for internal stuff

app/components/rdf-input-fields/rijksregister-input.hbs
Rahien, line 1: can't remove these in this pr yet i think?

app/utils/register-form-fields.js
```diff
@@ -5,8 +5,6 @@ import RDFAddressSelectorComponent from 'frontend-lmb/components/rdf-input-field
 import RDFPersonSelectorComponent from 'frontend-lmb/components/rdf-input-fields/person-selector';
 import RDFMandatarisFractieSelectorComponent from 'frontend-lmb/components/rdf-input-fields/mandataris-fractie-selector';
 import RDFMandatarisMandaatSelectorComponent from 'frontend-lmb/components/rdf-input-fields/mandataris-mandaat-selector';
-import RDFRijksRegisterInput from 'frontend-lmb/components/rdf-input-fields/rijksregister-input';
```
Rahien, line 8: need to keep this one around for now

app/components/rdf-input-fields/mandataris-fractie-selector.js
```diff
@@ -68,42 +69,37 @@ export default class MandatarisFractieSelector extends InputFieldComponent {
...
+  // TODO Maybe turn into task?
   @action
   async onSelectFractie(fractie) {
     this.updating = true;
-    const newMembership = this.store.createRecord('lidmaatschap', {
-      binnenFractie: fractie,
-    });
-    await newMembership.save();
-    this.membership = newMembership;
+    const uri = fractie.uri;
```
Rahien, line 88: i think this will break when you clear the input. we should check if it's null
```suggestion
    const uri = fractie?.uri;
```

Rahien (changes requested):

app/components/rdf-input-fields/mandataris-fractie-selector.js
```diff
@@ -39,6 +39,7 @@ export default class MandatarisFractieSelector extends InputFieldComponent {
     return this.args.field?.label || 'Fractie';
   }
 
+  // TODO unused
   get displayValue() {
```
ruben-bottu, line 43: Unused method

app/components/rdf-input-fields/mandataris-fractie-selector.js
```diff
@@ -68,42 +69,37 @@ export default class MandatarisFractieSelector extends InputFieldComponent {
...
+    const matches = await this.store.query('fractie', {
+      'filter[:uri:]': fractieUri,
     });
-    this.membership = matches.at(0);
+    this.selectedFractie = matches.at(0);
   }
 
+  // TODO Maybe turn into task?
   @action
   async onSelectFractie(fractie) {
```
ruben-bottu, line 86: Maybe should be a task? You get `this.updating = true` at the start and `this.updating = false` at the end for free with tasks

app/models/fractie.js
```diff
@@ -1,6 +1,7 @@
 import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
 
 export default class FractieModel extends Model {
+  @attr uri;
```
ruben-bottu, line 4: Otherwise Fractie URI is not loaded into frontend

app/components/rdf-input-fields/mandataris-fractie-selector.js
Rahien: this is true. If you feel it's worth the effort you can ofc do this

app/models/fractie.js
Rahien: this is a bit of a general gotcha :(

Rahien (approved):

PR #160, opened 2024-04-09
add some error handling to addressregister service

bfidlers:


app/services/addressregister.js
```diff
@@ -79,11 +79,18 @@ export default class AddressregisterService extends Service {
...
-      const results = await (
-        await fetch(
-          `/adressenregister/match?municipality=${suggestion.municipality}&zipcode=${suggestion.zipCode}&thoroughfarename=${suggestion.street}&housenumber=${suggestion.housenumber}`
-        )
-      ).json();
+      const response = await fetch(
+        `/adressenregister/match?municipality=${suggestion.municipality}&zipcode=${suggestion.zipCode}&thoroughfarename=${suggestion.street}&housenumber=${suggestion.housenumber}`
+      );
+      if (!response.ok) {
+        return addresses;
```
ruben-bottu, line 86: I was a bit confused by this a first, but I assume that you just want to return an emty list. It might be clearer to just `return []` here and on line 92. Your choice

PR #165, opened 2024-04-09
LMB-238: Show bestuursorgaan names next to the role

JonasVanHoof:
LMB-23

## Test
1. Go to the person page where you can see all the madatarissen of that person
2. Next to the role all the bestuursorgaan names shoulw be visible
3. When nog bestuursorgaan names only the role will be shown

image

### DIFF
--- a/app/components/mandatarissen/mandataris-card.hbs
+++ b/app/components/mandatarissen/mandataris-card.hbs
@@ -1,13 +1,8 @@
 <AuCard as |c|>
   <c.header>
     <div class="au-u-flex">
-      <AuHeading
-        @level="2"
-        @skin="4"
-        class="flex-grow"
-        {{did-insert (perform this.setBestuursOrganenForMandataris)}}
-      >
-        {{this.rolWithBestuurOrgaanNames}}
+      <AuHeading @level="2" @skin="4" class="flex-grow">
+        {{this.rol}}
       </AuHeading>
       <AuPill
         @skin="link"
--- a/app/components/mandatarissen/mandataris-card.hbs
+++ b/app/components/mandatarissen/mandataris-card.hbs
@@ -1,8 +1,15 @@
 <AuCard as |c|>
   <c.header>
     <div class="au-u-flex">
-      <AuHeading @level="2" @skin="4" class="flex-grow">
-        {{this.rol}}{{await this.bestuursOrgaanLabel}}
+      <AuHeading
+        @level="2"
+        @skin="4"
+        class="flex-grow"
+        {{did-insert (perform this.setBestuursOrganenForMandataris)}}
+      >
+        {{this.rol}}
+        -
+        {{this.bestuursOrgaanNames}}
       </AuHeading>
       <AuPill
         @skin="link"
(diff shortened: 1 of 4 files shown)


app/components/mandatarissen/mandataris-card.hbs
```diff
@@ -1,8 +1,13 @@
...
     <div class="au-u-flex">
-      <AuHeading @level="2" @skin="4" class="flex-grow">
-        {{this.rol}}
+      <AuHeading
+        @level="2"
+        @skin="4"
+        class="flex-grow"
+        {{did-insert (perform this.setBestuursOrganenForMandataris)}}
+      >
+        {{this.rolWithBestuurOrgaanNames}}
```
JonasVanHoof, line 10: Role was only used once and depending on the bestuursorgaan names we had to add or remove the dash inbetween so changed it to one getter for both. I simulated it here on how it would look if a user has to wait for 2000ms on the bestuursorgaan names.

@Rahien do we need the loader state? I think it can be shown like in the demo. All data without async is shown on init and the names will come behind. If we do want to show the user that something could be happening behind the name i suggest only a loader for the names but that's maybe a bit silly ;p


https://github.com/lblod/frontend-lokaal-mandatenbeheer/assets/36266973/f509931e-9a95-40d7-803b-205ba369b1f4

app/components/mandatarissen/mandataris-summary.js
```diff
@@ -27,8 +29,46 @@ export default class MandatenbeheerMandatarisSummaryComponent extends Component
     });
   }
 
+  get rolWithBestuurOrgaanNames() {
+    const rol = this.args.mandataris.bekleedt
+      .get('bestuursfunctie')
+      .get('label');
+
+    if (!this.bestuursOrgaanNames || this.bestuursOrgaanNames == '') {
```
ruben-bottu, line 37: I would always use triple equals `===`

app/components/mandatarissen/mandataris-summary.hbs
```diff
@@ -1,7 +1,7 @@
 <div class="au-o-grid au-o-grid--flush">
   <div class="au-o-grid__item au-u-3-5">
-    <p>
-      <strong>{{this.rol}}</strong>
+    <p {{did-insert (perform this.setBestuursOrganenForMandataris)}}>
```
ruben-bottu, line 3: Cool workaround!

ruben-bottu (approved):

JonasVanHoof:
@ruben-bottu  I was checking for the change of the triple equals by creating a new mandataat (without a mandaat). Than it errored on the the get rol method
```js
  get rolWithBestuurOrgaanNames() {
    const bestuursfunctie =
      this.args.mandataris.bekleedt.get('bestuursfunctie');

    if (!bestuursfunctie) {
      return 'ONBEKEND';
    }

    const rol = bestuursfunctie.get('label') ?? '';

    if (!this.bestuursOrgaanNames || this.bestuursOrgaanNames === '') {
      return rol;
    }
    return `${rol} - ${this.bestuursOrgaanNames}`;
  }
  ```
  
  Because the error occured I wanted to change it to the code above but it will only fix the person page. When going to the details of the mandaat it errors on the id.
    
image

  
  Should we make the mandaat field required in the form and iof the user cannot create a mandaat that the button is disabled on the person page.
  
image

PR #166, opened 2024-04-10
Ben/lmb 227 refactor onafhankelijke fractie

bfidlers:
## Description
This PR solves a few different problems, if no onafhankelijke fractie exists, one is created when opening a form that contains a fractie selector. When selecting fracties, only one fractie that is of the onafhankelijke fractietype is shown. The correct bestuursorganen-in-de-tijd are provided to the component, so this new fractie can be attached to the correct bestuursorganen in de tijd.

## Testing
Testing can happen in a few different places, Gemeente Aarschot is a good option, since it does not contain a fractie with the name or type onafhankelijk

1. Go to the fractie route under bestuursorganen, to check which fracties actually exist
2. Go to a mandataris and open the corrigeer fouten or update state form (best to check both)
3. If you go to the fractie-selector you should see a new onafhankelijke fractie (if you look in the ember debugger under data, you can see this fractie is actually of the type onafhankelijk)
4. If you go back to the fractie route you see this fractie is added here as well

If you want to check the fact that only one fractie of the onafhankelijke fractietype is shown, you should first test this and afterwards run the migration included in this branch ben/lmb-369-migration-onafhankelijke-fracties this would result in two fracties of the onafhankelijke type in e.g Gemeente Aalst, which you can check in the fractie route, but which is not shown in the mandataris form.


app/components/mandatarissen/fractie-selector.js
```diff
@@ -40,34 +55,35 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
...
   select(fractie) {
     this._fractie = fractie;
     this.args.onSelect(fractie);
   }
+
+  search = task({ restartable: true }, async (searchData) => {
+    await timeout(SEARCH_TIMEOUT);
+    let searchResults = await this.fetchFracties(searchData);
+    return searchResults;
+  });
```
JonasVanHoof, line 88: Nitpick: would change it to:
```js
search = restartableTask( async (searchData) => {
   ...
})
```

app/components/mandatarissen/fractie-selector.js
```diff
@@ -40,34 +55,35 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
...
   }
 
   @action
   select(fractie) {
     this._fractie = fractie;
     this.args.onSelect(fractie);
   }
+
+  search = task({ restartable: true }, async (searchData) => {
+    await timeout(SEARCH_TIMEOUT);
```
JonasVanHoof, line 85: There is also a timeout on L45 is it needed 1000ms inbetween in the search?

app/components/rdf-input-fields/mandataris-fractie-selector.js
```diff
@@ -40,27 +36,48 @@ export default class MandatarisFractieSelector extends InputFieldComponent {
...
+  async loadBestuursorganen() {
+    this.bestuurseenheid = this.currentSession.group;
+    const queryParams = this.router.currentRoute.queryParams;
+
+    const bestuursorganen = await this.store.query('bestuursorgaan', {
+      'filter[bestuurseenheid][id]': this.bestuurseenheid.id,
+      'filter[:has-no:deactivated-at]': true,
+      'filter[:has-no:is-tijdsspecialisatie-van]': true,
+      'filter[classificatie][id]':
+        this.decretaleOrganen.classificatieIds.join(','), // only organs with a political mandate
```
JonasVanHoof, line 52: this.decretaleOrganen is undefined: bug?

bfidlers: fixed it, forgot to inject the service

Rahien (changes requested):
Looks good and seems to work as it should except for the fractie selector in forms where the currently selected fraction isn't filled in

Rahien:
> Looks good and seems to work as it should except for the fractie selector in forms where the currently selected fraction isn't filled in

Must have been on the wrong branch on the backend side. Works.

Rahien (approved):

PR #169, opened 2024-04-10
LMB-294: Forbid mandatees from changing fraction

JonasVanHoof:
## Description 
When a fraction is added to a mandataris the user can only change the fraction to one it had before or to independent. This limitation can only be in the update mandataris and NOT in the correct mandataris.

## Test

1. Go to a mandataris or create a new one without a fraction
2. Select a fraction (all options should be in the list)
3. Update it again. The fraction list should contain the previous option and independent

### DIFF
--- a/app/components/mandatarissen/fractie-selector.js
+++ b/app/components/mandatarissen/fractie-selector.js
@@ -39,7 +39,7 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
       fracties = [...fracties, onafhankelijke];
     }
 
-    if (this.isChangingFraction && this.mandataris) {
+    if (this.isChangingFraction && this.mandataris && this._fractie) {
       if (await this.isFractionIndependent(this._fractie)) {
         const mandataries = await this.store.query('mandataris', {
           include: 'heeft-lidmaatschap,heeft-lidmaatschap.binnen-fractie',
@@ -54,7 +54,14 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
         const fractionsForMandataries = [];
         for (const mandate of mandataries) {
           const lidmaatschap = await mandate.heeftLidmaatschap;
+          if (!lidmaatschap) {
+            continue;
+          }
           const fraction = await lidmaatschap.binnenFractie;
+          if (!fraction) {
+            continue;
+          }
+
           if (
             !fractionsForMandataries.find(
               (fractionModel) => fractionModel.id == fraction.id
--- a/app/components/mandatarissen/fractie-selector.js
+++ b/app/components/mandatarissen/fractie-selector.js
@@ -39,16 +39,38 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
       fracties = [...fracties, onafhankelijke];
     }
 
-    if (this.isChangingFraction) {
+    if (this.isChangingFraction && this.mandataris) {
       if (await this.isFractionIndependent(this._fractie)) {
-        const previousFraction = null;
-        console.log(`set previous fraction`, previousFraction);
+        const mandataries = await this.store.query('mandataris', {
+          include: 'heeft-lidmaatschap,heeft-lidmaatschap.binnen-fractie',
+          filter: {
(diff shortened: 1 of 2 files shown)


JonasVanHoof:
The logic for limiting the options in the fraction selector when "Wijzig huidige situatie"  is added.
image


**Blocked by:** https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/166/files#r1560504829

app/components/mandatarissen/fractie-selector.js
```diff
@@ -29,18 +29,64 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
...
-      fracties = [...fracties, onafhankelijke];
+      this.fractieOptions = [...fracties, onafhankelijke];
+      return;
     }
+
     this.fractieOptions = fracties;
   }
 
+  async getFractiesWhenUpdateState() {
+    const mandataries = await this.store.query('mandataris', {
```
bfidlers, line 52: I find mandataries really confusing. Maybe better to use mandatarissen or maybe mandatees?

app/components/mandatarissen/fractie-selector.js
```diff
@@ -85,4 +131,38 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
...
+  async getIndependentFractie(fracties) {
+    for (const fractie of fracties) {
+      const isIndependent = await this.isFractieIndependent(fractie);
+
+      if (isIndependent) {
+        return fractie;
+      }
+    }
+    return null;
+  }
```
bfidlers, line 154: Wondering why you added this? This replaces practically a oneliner, but I don't feel like this really makes it more readable, so I don't really get why. Or was it broken before?

app/components/mandatarissen/fractie-selector.js
```diff
@@ -29,18 +29,64 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
   }
 
   async loadFracties() {
-    let fracties = await this.fetchFracties();
+    let fracties = [];
+
+    if (this.isUpdatingFractie && this._fractie) {
```
bfidlers, line 34: Do you need a dedicated function for this? Can't you just use this.args.isUpdatingFractie? Also might be better to use another variable name, since isUpdatingFractie is a bit vague, in the other use case, you also update the fractie ... maybe better something like "isUpdatingState"?

app/components/mandatarissen/fractie-selector.js
```diff
@@ -85,4 +131,38 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
...
+      const isIndependent = await this.isFractieIndependent(fractie);
+
+      if (isIndependent) {
+        return fractie;
+      }
+    }
+    return null;
+  }
+
+  get mandataris() {
```
bfidlers, line 156: Personally I wouldn't do this, because it seems pretty redundant, but I can live with it.

bfidlers (changes requested):
Left a few comments. 
I would update the variable names since these are pretty confusing. The other comments are more personal preference, so that's your call.
Functionally everything seems fine.

app/components/mandatarissen/fractie-selector.js
JonasVanHoof: Updated the naming

app/components/mandatarissen/fractie-selector.js
JonasVanHoof: You are right the oneliner is better. I think while implementing I used it on multiple places thats why it moved to a spearte method, good one

bfidlers (approved):
Thanks, looks good!

app/components/mandatarissen/fractie-selector.js
JonasVanHoof: Updated the args value that's indeed some code thats not needed there. 👌🏻 
Hm I thought about to but when checking the actions and button names I came out on `isUpdatingFractie` becasue the other action is `correcting` the fractie. Just using the `isUpdatingState` name will not be enough because the component is always updating the fractie ;p

Let's see on monday what the naming says :D

PR #170, opened 2024-04-11
LMB-236: Navigate to persoon from mandataris

JonasVanHoof:
LMB-236

## Description

The user was not able to go back to the mandates of the person itself. Now they can click the name at the top of the detailed mandataris to go back to the persons mandatarissen overview.

## Test
1. Go to a persons mandatarissen
2. Open the detail of a mandaat of the person
3. Now you will see that the name of the person is a link and that you can go back to the persons overview by clicking it

image


ruben-bottu (approved):
Looks great! This was indeed a bit of a pain point in the navigation

PR #173, opened 2024-04-15
LMB- 370: dct:modified to mandataris, fractie and person

JonasVanHoof:
These entities should have a property dct:modified to keep track of the changes.

## Test
When checking out this PR make sure to also checkout these branches
* **App:** https://github.com/lblod/app-lokaal-mandatenbeheer/pull/115 
* **Form-content-service:** https://github.com/lblod/form-content-service/pull/44

### DIFF
--- a/app/models/fractie.js
+++ b/app/models/fractie.js
@@ -4,6 +4,7 @@ export default class FractieModel extends Model {
   @attr uri;
   @attr naam;
   @attr('uri-set') generatedFrom;
+  @attr('datetime') modified;
 
   @belongsTo('fractietype', {
     async: true,
--- a/app/models/fractie.js
+++ b/app/models/fractie.js
@@ -4,7 +4,12 @@ export default class FractieModel extends Model {
   @attr uri;
   @attr naam;
   @attr('uri-set') generatedFrom;
-  @attr('datetime') modified;
+  @attr('datetime', {
+    defaultValue() {
+      return new Date();
+    },
+  })
+  modified;
 
   @belongsTo('fractietype', {
     async: true,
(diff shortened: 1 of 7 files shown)


app/models/fractie.js
```diff
@@ -4,6 +4,7 @@ export default class FractieModel extends Model {
   @attr uri;
   @attr naam;
   @attr('uri-set') generatedFrom;
+  @attr('datetime') modified;
```
JonasVanHoof, line 7: Should we add a default? 
```js
  @attr('datetime', {
    defaultValue() {
      return new Date();
    },
  })
  modified;

File: app/routes/organen/fracties.js [hunk 2]
```diff
@@ -102,7 +102,7 @@ export default class FractiesRoute extends Route {
   }
 
   @action
-  reloadModel() {
+  async reloadModel() {
```
ruben-bottu, line 105: Not sure how useful the async is here if you don't await anything.

ruben-bottu (approved):

ruben-bottu (commented):
Should there maybe also be a dct:modified for creating instances?

ruben-bottu (changes requested):
This feedback applies to app, form-content and frontend:

Newly created instances should also have a dct:modified. This way our data will be more easy to consume for the vendors.

app/routes/organen/fracties.js
JonasVanHoof: Removed it

JonasVanHoof:
> This feedback applies to app, form-content and frontend:
> 
> Newly created instances should also have a dct:modified. This way our data will be more easy to consume for the vendors.

When an entity is created through the store it gets the default value for modified (new Date() at that point in time) but indeed when a post is done through the form-content service this could not be the case. Will check this out. 

Next to the upcoming change in the form-content-service I think we can already merge the current PR's right?

ruben-bottu (approved):
Approved on the condition that form-content-service also creates dct:modified for new entities

PR #175, opened 2024-04-17
Add Mandataris Publication Status

ruben-bottu:
De "Bekrachtig" button is replaced by a publication status selector for more granularity.
image

This can be tested by creating a new mandate and checking if its status is set to draft. All mandates without a publication status should default to "bekrachtigd".

Selecting a different publication status should immediately persist the new status and it should survive a refresh.


app/components/mandatarissen/mandataris/publication-status-selector.hbs
bfidlers, line 1: Would suggest adding a disabled option here, in the case the mandataris is already bekrachtigd.

app/components/mandatarissen/mandataris-edit-prompt.hbs
bfidlers, line 1: UI wise this needs to be reviewed, but is good for now.

app/components/mandatarissen/mandataris-edit-prompt.js
```diff
@@ -78,10 +102,10 @@ export default class MandatenbeheerMandatarisEditPromptComponent extends Compone
...
 
   @action
-  async onBekrachtig() {
-    this.editMode = null;
-    this.args.mandataris.isDraft = false;
-    this.args.mandataris.save();
+  async onUpdatePublicationStatus(publicationStatus) {
+    this.mandataris.publicationStatus = publicationStatus;
+    await this.mandataris.save();
+    this.router.refresh();
```
bfidlers, line 108: This refreshing is not that user friendly

app/utils/get-draft-status.js
bfidlers, line 1: Bit inconsistent that you have this util function for draft state, but not for the bekrachtigd state, but ...

bfidlers (changes requested):
Few comments here and there that should be looked at, also the update-state is broken for me, if you try to change a mandataris that was already there or a new one as well. 

Screenshot 2024-04-22 at 12 03 24

app/utils/get-draft-status.js
ruben-bottu: That is because getDraftStatus is used in multiple files in the codebase, and getBekrachtigdStatus is not. Once getBekrachtigdStatus is needed in multiple files, it can still be turned into a helper function, but for now it makes the most sense to me to keep it where it is used.

app/components/mandatarissen/update-state.js
```diff
@@ -162,14 +163,14 @@ export default class MandatarissenUpdateState extends Component {
...
       rangorde: this.rangorde,
-      isDraft: true,
       start: this.date,
-      status: this.newStatus,
       einde: endDate,
+      bekleedt: this.args.mandataris.bekleedt,
+      isBestuurlijkeAliasVan: this.args.mandataris.isBestuurlijkeAliasVan,
+      beleidsdomein: this.selectedBeleidsdomeinen,
+      status: this.newStatus,
+      publicationStatus: await getDraftStatus(this.store),
```
bfidlers, line 173: I think we need to check this business rule, because it seems weird to me that if you do an update state, your new mandataris becomes draft, this doesn't really seems to make sense, because draft is meant for input of an installatievergadering, but I suppose if you do update state the installatievergadering already passed.

app/components/mandatarissen/mandataris-edit-prompt.js
```diff
@@ -24,8 +28,47 @@ export default class MandatenbeheerMandatarisEditPromptComponent extends Compone
...
+  }
+
+  async setBekrachtigdStatus() {
+    const bekrachtigdStatus = await this.getBekrachtigdStatus();
+
+    this.disabled = true;
+    this.selectedPublicationStatus = bekrachtigdStatus;
+  }
+
+  async loadPublicationStatusOptions() {
```
bfidlers, line 58: This results in some weird behaviour when you update state. Then the old publicationStatusOptions are shown, might be better to fetch this in the component meant for the publicatieStatus, then you might be able to just invalidate the contents of the component instead of the whole route?

bfidlers (commented):
Still one problem with reloading, and a business rule we need to check, but that can wait. Otherwise looks good.

bfidlers (approved):
Looks good!

PR #177, opened 2024-04-22
Spaces allowed again in update rangorde mandataris

bfidlers:
Screenshot 2024-04-22 at 08 59 32


JonasVanHoof (approved):
Ran it. Works! 

Question: Does the backend trim the spaces after the text?

PR #186, opened 2024-04-25
LMB-388: Tabullar component for draft mandataris instances in orgaan

JonasVanHoof:
## Description

The current table in the `orgaan/mandatarissen` route looks like this.
image

We created a new component because we want that the user is able to edit the draft mandatarissen. We add configuration for showing columns. This because we want to use this component multiple times but not with all columns always visible.
All links to persons or mandatarissen are also removed. 
image

## How to test

1. Go to the corresponding link http://localhost:4200/organen/<ID>/mandatarissen and see if the tables contains all the columns as before (voornaam, achternaam, fracties, mandaat, start- en eind datum)
2. You are not routed to the person or mandaat when clicking on it
3. You do not have pagination


app/routes/organen/orgaan/mandatarissen.js
```diff
@@ -161,12 +161,13 @@ export default class OrganenMandatarissenRoute extends Route {
   }
 
   getOptions(params, bestuursOrgaan) {
+    const NO_PAGINTATION = {
+      number: params.page,
+      size: 9999,
+    };
```
JonasVanHoof, line 167: Thought that this was the clearest way to say that it does not have pagination

app/styles/app.scss
```diff
@@ -7,6 +7,7 @@
 @import "project/c-document-status-pills";
 @import "project/c-sidebar-container";
 @import "project/c-status-label";
+@import "project/c-skeleton";
```
JonasVanHoof, line 10: Ah should remove this again!

app/styles/project/_c-skeleton.scss
```diff
@@ -0,0 +1,17 @@
...
+  0% {
+    background-color: var(--au-white);
+  }
+  50% {
+    background-color: var(--au-gray-100);
+  }
+  100% {
+    background-color: #f1f3f6;
+  }
+}
```
JonasVanHoof, line 17: Just like this remove it

app/routes/organen/orgaan/mandatarissen.js
JonasVanHoof: Maybe we also should set the page number to always be 0?

bfidlers (commented):
Looks good, nice table, but I think it is in the wrong place, I think it was supposed to go to the new route for preparing the installatievergadering (verkiezingen/verkiezingsuitslag/prepare)

app/routes/organen/orgaan/mandatarissen.js
bfidlers: Small typo here as well.

app/templates/organen/orgaan/mandatarissen.hbs
```diff
@@ -25,13 +25,21 @@
...
-    @mandaatRoute="mandatarissen.mandataris"
+    @title=""
+    @showVoornaam={{true}}
+    @showNaam={{true}}
+    @showFractie={{true}}
+    @showFunctie={{true}}
+    @showRangorde={{false}}
+    @showBevoegdheid={{false}}
+    @showStartMandaat={{true}}
+    @showEindeMandaat={{true}}
```
bfidlers, line 42: I think it is nice have settings for which column you show, but I think some colums should always be shown, no?

app/routes/organen/orgaan/mandatarissen.js
JonasVanHoof: Removed it in this route

app/templates/organen/orgaan/mandatarissen.hbs
JonasVanHoof: Yes indeed! I was thinking at a 'showAll' & 'showPersonInfo' & de andere dan apart ?

JonasVanHoof: Why: When looking at the mockups the dates or not in there and the person info is always visible

app/components/verkiezingen/draft-mandataris-list.hbs
```diff
@@ -0,0 +1,101 @@
+<div class="au-c-body-container au-c-body-container--scroll">
```
bfidlers, line 1: I think it is a bit dangerous to put this in the component, I would put this in the template, then we have more control in the template if we want to put the scroll on the page or on the table.

app/routes/verkiezingen/verkiezingsuitslag/prepare.js
```diff
@@ -1,9 +1,60 @@
...
-  async model() {
+  @service store;
+
+  queryParams = {
+    filter: { refreshModel: true },
+    sort: { refreshModel: true },
+  };
+
+  async model(params) {
+    const verkiezingsuitslag = this.modelFor('verkiezingen.verkiezingsuitslag');
```
bfidlers, line 14: Personally I would change this to parentModel, this is what we mostly use in other routes and then it is clear that this can contain multiple resources.

app/routes/verkiezingen/verkiezingsuitslag/prepare.js
```diff
@@ -1,9 +1,60 @@
...
+    }
+
     return {
-      ...this.modelFor('verkiezingen.verkiezingsuitslag'),
+      ...verkiezingsuitslag,
+      mandatarissen: mandatarissen,
     };
   }
+
+  getOptions(params, bestuursOrgaan) {
```
bfidlers, line 32: This needs to be reviewed when we construct the right tables, but for now this is good for testing the table functionalities.

bfidlers (approved):
Few small remarks, but looks good otherwise.

PR #187, opened 2024-04-25
LMB-415: modied store.save is overwritting the form changes of form-content-service

JonasVanHoof:
## Description

When correcting a mandataris the changes are saved but than overwritten by the save of ember. This because the dct:modified was saved in the ember code.

## How to test

Modified should be correct when:
1. create new mandaat
2. wijzig een istuatie
3. Corrigeer situatie


Rahien (approved):
this is an important fix but we don't always store a dct modified on our ember models yet. Still it's important that this fix moves through because otherwise things can break hard, so approving. Will have a look and see if i can auto update dct:modfied another way

PR #190, opened 2024-04-26
Ben/lmb 410 fix bestuursperiod selector

bfidlers:
## Description

In the bestuursperiod selector only the bestuursperiods of one bestuursorgaan where shown, now all bestuursperiods are shown.

## How to test

Create a new bestuursperiod on a bestuursorgaan and check that this bestuursperiod actually show up in a bestuursperiod-selector. This can be validated with the bestuurperiod selector in the fracties or in the organen route.


Rahien (approved):
This is a bit ugly but that's because the domain model is bad here. Approving since stories exist to make bestuursperiod a true concept so we can get rid of this logic after

PR #193, opened 2024-04-29
Add Aangewezen Burgemeester Status Code

ruben-bottu:
## Description

The list of mandataris statusses is extended with a new 'aangewezen' status. Selecting this status should only be possible when selecting the status of a Burgemeester mandataris.

## How to test

- Go to the mandaat "Wijzig huidige situatie" modal. If the mandataris is a Burgemeester, "Aangewezen" should be an option in the status dropdown. Otherwise "Aangewezen" should not be in the dropdown.
- Go to the mandaat "Corrigeer fouten" modal. Here the "Aangewezen" status should also only appear when the mandataris is a Burgemeester.

image

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/124

### DIFF
--- a/app/components/rdf-input-fields/mandataris-status-selector.js
+++ b/app/components/rdf-input-fields/mandataris-status-selector.js
@@ -1,0 +1,23 @@
+import { queryRecord } from 'frontend-lmb/utils/query-record';
+import ConceptSchemeSelectorComponent from './concept-selector';
+import { MANDATARIS_AANGEWEZEN_STATE } from 'frontend-lmb/utils/well-known-uris';
+
+export default class RdfInputFieldsMandatarisStatusSelectorComponent extends ConceptSchemeSelectorComponent {
+  async fetchOptions(searchData) {
+    const statuses = await super.fetchOptions(searchData);
+    const mandatarisUri = this.storeOptions.sourceNode.value;
+    const mandataris = await queryRecord(this.store, 'mandataris', {
+      'filter[:uri:]': mandatarisUri,
+    });
+    const isBurgemeester = (await (await mandataris.bekleedt).bestuursfunctie)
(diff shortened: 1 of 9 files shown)


app/components/rdf-input-fields/mandataris-status-selector.js
```diff
@@ -0,0 +1,23 @@
+import { queryRecord } from 'frontend-lmb/utils/query-record';
+import ConceptSchemeSelectorComponent from './concept-selector';
+import { MANDATARIS_AANGEWEZEN_STATE } from 'frontend-lmb/utils/well-known-uris';
+
+export default class RdfInputFieldsMandatarisStatusSelectorComponent extends ConceptSchemeSelectorComponent {
+  async fetchOptions(searchData) {
+    const statuses = await super.fetchOptions(searchData);
+    const mandatarisUri = this.storeOptions.sourceNode.value;
```
Rahien, line 8: this won't work if they change both the mandate and the status. You should fetch the current mandate from the form data and check if it has the burgemeester classification

app/components/rdf-input-fields/mandataris-status-selector.js
```diff
@@ -0,0 +1,23 @@
...
+      'filter[:uri:]': mandatarisUri,
+    });
+    const isBurgemeester = (await (await mandataris.bekleedt).bestuursfunctie)
+      .isBurgemeester;
+
+    if (isBurgemeester) {
+      return statuses;
+    }
+
+    return statuses.filter(
```
Rahien, line 19: I tested and somehow the filtering also doesn't work for normal members of the council

Rahien (changes requested):

Rahien (approved):

PR #195, opened 2024-04-30
LMB-391: Change rangorde up and down of mandataris in orgaan

JonasVanHoof:
## Description

We want to change the rangorde of a mandataris in orgaan to be moved up or down in the table. The input field is a free text input but we are setting some limitations on it warning users of potential issues.

* We support changing the rangorde using buttons from 1 to 20 inclusive
* Gibberish text is allowed but shows a warning
* A rangorde is automatically generated based on the mandate of the mandataris when clicking plus or mandataris
* Sorting works based on the numeric value of the rangorde and is unchanged for other columns
* when changing the rangorde using the buttons, the mandataris switches rangorde with the mandataris that previously had that rangorde
* changing rangorde also takes into account the sorting

## How to test

Check the rangorde of the mandaten

## Attachments

image

### DIFF
--- a/app/components/verkiezingen/rangorde-input.js
+++ b/app/components/verkiezingen/rangorde-input.js
@@ -100,7 +100,7 @@ export default class VerkiezingenRangordeInputComponent extends Component {
   }
 
   get isPlusDisabled() {
-    return !this.rangordeInteger;
+    return !this.rangordeInteger || this.rangordeInteger >= 20;
   }
 
   get rangordeAsStringMapping() {
--- a/app/components/verkiezingen/rangorde-input.js
+++ b/app/components/verkiezingen/rangorde-input.js
@@ -1,89 +1,67 @@
 import Component from '@glimmer/component';
 
 import { action } from '@ember/object';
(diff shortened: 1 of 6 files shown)


Rahien (changes requested):
need to be able to enter other things than numbers.

ruben-bottu:
Functionality wise looks good

app/components/verkiezingen/rangorde-input.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  findOrderInString(possibleString) {
+    if (!possibleString) {
+      return null;
+    }
+    let foundNumber = null;
+    Object.keys(rangordeStringMapping).forEach((key) => {
+      if (possibleString.startsWith(key)) {
+        foundNumber = rangordeStringMapping[key];
```
ruben-bottu, line 65: Maybe the loop should be ended early if a match is found?

ruben-bottu (approved):
Looks good

Rahien (approved):

PR #199, opened 2024-05-03
Add Mandataris To Orgaan in Voorbereiding Legislatuur

ruben-bottu:
## Description

You should be able to add mandataris instances to an orgaan in "Voorbereiding legislatuur"

## How to test

- Go to the voorbereiding legislatuur page
- Click on the "Toevoegen" button under the "Voorbereiding legislatuur" table
- Add a mandataris and see it appear in the table

### DIFF
--- a/app/controllers/verkiezingen/verkiezingsuitslag/prepare.js
+++ b/app/controllers/verkiezingen/verkiezingsuitslag/prepare.js
@@ -32,7 +32,6 @@ export default class PrepareInstallatievergaderingController extends Controller 
   async onCreate({ instanceTtl, instanceId }) {
     this.editMode = null;
     await syncNewMandatarisMembership(this.store, instanceTtl, instanceId);
-    // TODO not sure why the timeout is needed
     setTimeout(() => this.router.refresh(), 1000);
   }
 
--- a/app/controllers/verkiezingen/verkiezingsuitslag/prepare.js
+++ b/app/controllers/verkiezingen/verkiezingsuitslag/prepare.js
@@ -2,11 +2,9 @@ import Controller from '@ember/controller';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { service } from '@ember/service';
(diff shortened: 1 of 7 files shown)


app/controllers/verkiezingen/verkiezingsuitslag/prepare.js
```diff
@@ -1,7 +1,51 @@
...
+  cancel() {
+    this.editMode = null;
+  }
+
+  @action
+  async onCreate({ instanceTtl, instanceId }) {
+    this.editMode = null;
+    await syncNewMandatarisMembership(this.store, instanceTtl, instanceId);
+    // TODO not sure why the timeout is needed
+    setTimeout(() => this.router.refresh(), 1000);
```
ruben-bottu, line 36: Not sure why the timeout is needed

Rahien: It's because of the resources cache maybe not being updated when the form-content-service returns. This extra second makes it much more likely that this happened. (this is because that is done using deltas and the content service doesn't know whether they have been handled so it can't wait for it)

Rahien (approved):
Don't forget to remove that TODO :)

app/routes/verkiezingen/verkiezingsuitslag/prepare.js
```diff
@@ -15,21 +17,27 @@ export default class PrepareInstallatievergaderingRoute extends Route {
...
-      mandatarissen = await this.getMandatarissen(params, bestuursorgaan);
+    if (bestuursorgaanInTijd) {
+      mandatarissen = await this.getMandatarissen(params, bestuursorgaanInTijd);
     }
 
-    let kandidatenlijsten = await this.getKandidatenLijsten(bestuursorgaan);
+    let kandidatenlijsten =
+      await this.getKandidatenLijsten(bestuursorgaanInTijd);
+
+    const bestuursorgaan = bestuursorgaanInTijd.isTijdsspecialisatieVan;
```
bfidlers, line 31: is not used I think

PR #203, opened 2024-05-07
Ben/lmb 392 delete mandataris

bfidlers:
## Description

Add a delete button to the draft mandataris table in prepare installatievergadering. This creates a thombstone.
Screenshot 2024-05-07 at 10 00 10

## How to test

Go to the prepare installatievergadering route and delete a mandataris. You should go to the gemeente Aalst and select the old bestuursorgaan in tijd, otherwise, you probably won't have mandatarissen in the table and you'll first have to add some (if you have ran the latest migrations). 
If you want to check if the resources were actually deleted, you can check which resources are in the database and this should show something like this for the deleted mandataris:
Screenshot 2024-05-07 at 09 32 06

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/132
- https://github.com/lblod/mandataris-service/pull/1

### DIFF
--- a/app/components/verkiezingen/draft-mandataris-list.js
+++ b/app/components/verkiezingen/draft-mandataris-list.js
@@ -1,10 +1,14 @@
 import Component from '@glimmer/component';
 
+import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { orderMandatarissenByRangorde } from 'frontend-lmb/utils/rangorde';
+import { JSON_API_TYPE } from 'frontend-lmb/utils/constants';
 
 export default class DraftMandatarisListComponent extends Component {
+  @service toaster;
+
   @tracked isModalOpen = false;
   @tracked mandataris;
   @tracked editBeleidsdomeinen;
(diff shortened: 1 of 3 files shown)


app/components/verkiezingen/draft-mandataris-list.js
```diff
@@ -65,4 +68,19 @@ export default class DraftMandatarisListComponent extends Component {
     await this.mandataris.save();
     this.closeModal();
   }
+
+  @action
+  async removeMandataris(mandataris) {
+    mandataris
+      .destroyRecord()
+      .then(() => {
```
ruben-bottu, line 76: Why the then catch syntax?

app/routes/verkiezingen/verkiezingsuitslag/prepare.js
```diff
@@ -59,6 +59,7 @@ export default class PrepareInstallatievergaderingRoute extends Route {
         'bekleedt.bestuursfunctie',
         'heeft-lidmaatschap.binnen-fractie',
         'status',
+        'beleidsdomein',
```
ruben-bottu, line 62: Why is this necessary?

ruben-bottu (approved):

app/components/verkiezingen/draft-mandataris-list.js
bfidlers: just felt like it :)

app/routes/verkiezingen/verkiezingsuitslag/prepare.js
bfidlers: to decrease number of calls to resources

PR #205, opened 2024-05-07
Lmb 420/entity to add email to bestuursorgaan

JonasVanHoof:
## Description

We want every bestuurseenheid to add an email adress as contact information. The make this possible anew page `/settings` is created where they can fill in a form. This form will update the bestuurseenheid-contact model with the inserted email adres.

## How to test

1. In the topbar you find the bestuurseenheid name
2. This is a dropdown where the `Instellingen` is added
3. When routing through `/settings` a form is shown where the user can input the e-mailadres

## Notes
- now when you confirmed the changes in the form the button will show a loading message 
    -  the modal closes after it has saved the changes

## Attachments

image

image

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/133

### DIFF
--- a/app/components/form/instance.hbs
+++ b/app/components/form/instance.hbs
@@ -31,6 +31,7 @@
           >Pas aan</AuButton>
           <Form::SaveWithHistory
             @save={{this.saveInstance}}
+            @isSaving={{this.save.isRunning}}
             @updateValue={{this.updateHistoryMessage}}
             @isModalOpen={{this.isSaveHistoryModalOpen}}
           >Pas aan</Form::SaveWithHistory>
--- a/app/components/form/save-with-history.hbs
+++ b/app/components/form/save-with-history.hbs
@@ -17,7 +17,11 @@
   <Modal.Footer>
     <AuToolbar as |Group|>
       <Group>
-        <AuButton {{on "click" @save}}>
+        <AuButton
+          {{on "click" @save}}
+          @loading={{@isSaving}}
+          @loadingMessage="Aan het opslaan"
+        >
           Bevestig
         </AuButton>
       </Group>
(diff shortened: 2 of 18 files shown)


app/components/form/save-with-history.hbs
```diff
@@ -17,7 +17,11 @@
   <Modal.Footer>
     <AuToolbar as |Group|>
       <Group>
-        <AuButton {{on "click" @save}}>
+        <AuButton
+          {{on "click" @save}}
+          @loading={{@isSaving}}
+          @loadingMessage="Aan het opslaan"
+        >
```
JonasVanHoof, line 24: Not PR related but added this as when you are waiting on the save nothing is shown to the user

app/components/form/instance.hbs
```diff
@@ -31,6 +31,7 @@
           >Pas aan</AuButton>
           <Form::SaveWithHistory
             @save={{this.saveInstance}}
+            @isSaving={{this.save.isRunning}}
```
JonasVanHoof, line 34: To show the loading state of the button in the confirm modal

app/config/custom-inflector-rules.js
```diff
@@ -71,3 +71,4 @@ inflector.irregular(
 inflector.irregular('minister-position', 'minister-positions');
 
 inflector.irregular('nationality', 'nationalities');
+inflector.irregular('bestuurseenheid-contact', 'bestuurseenheid-contact');
```
Rahien, line 74: would not change this to singular

app/controllers/settings.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default class SettingsController extends Controller {
+  @service currentSession;
+  @tracked bestuurseenheidEmail;
+
+  get bestuurseenheid() {
+    return { email: this.bestuurseenheidEmail };
+  }
+
+  saveBestuurseenheidSettings = restartableTask(async () => {});
```
Rahien, line 15: this looks very strange why do we need it?

Rahien (changes requested):

app/controllers/settings.js
JonasVanHoof: Removed it completely, didn't know that you could leav out the argument of `onSave` in form::instance

Rahien (approved):

PR #206, opened 2024-05-08
Ben/lmb 422 bestuursperiode concept

bfidlers:
## Description

Add new bestuursperiode model, update current bestuursperiode selector and refactor the uses of the bestuursperiode util and tijdsspecialisaties service to take advantage of these new concepts.

## How to test

The bestuursperiode selector is used in three different places, all in the organen route (index, orgaan and fracties), be sure to check these. But these changes could have a big impact, so it is a good idea to also just click around a bit and iteract with different bestuursorganen in de tijd etc.

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/135


app/components/mandatarissen/update-state.js
```diff
@@ -52,19 +50,8 @@ export default class MandatarissenUpdateState extends Component {
...
-          startDate: moment(
-            this.bestuursorganenOfMandaat[0].bindingStart
-          ).format('YYYY-MM-DD'),
-          endDate: moment(this.bestuursorganenOfMandaat[0].bindingEinde).format(
-            'YYYY-MM-DD'
-          ),
-        }
-      );
+    this.bestuursperiode =
+      await this.bestuursorganenOfMandaat[0].heeftBestuursperiode;
```
JonasVanHoof, line 54: Do we need to catch if no bestuursperiode is found? First check if the item on index 0 is there?

app/controllers/organen/index.js
```diff
@@ -64,37 +55,22 @@ export default class OrganenIndexController extends Controller {
...
-    });
-    return bestuursorganen.firstObject;
-  }
-
-  async createDefaultBestuursorgaanInTijd(bestuursorgaan, bestuursperiod) {
+  async createDefaultBestuursorgaanInTijd(bestuursorgaan) {
+    // TODO look into if we really need these dates and what they should be.
+    const similarBestuursorgaanInTijd = (
+      await this.model.selectedPeriod.heeftBestuursorganenInTijd
+    )[0];
```
JonasVanHoof, line 68: Same here? Cheking on  index 0

app/services/bestuursperioden.js
```diff
@@ -0,0 +1,44 @@
+import Service from '@ember/service';
+import { inject as service } from '@ember/service';
```
JonasVanHoof, line 2: Inject as is not necessary anymore

JonasVanHoof (approved):
Looks good !

app/components/mandatarissen/update-state.js
bfidlers: Shouldn't really be possible, but can't hurt to check.

PR #207, opened 2024-05-10
Lmb 388/system notification resource

JonasVanHoof:
## Description

Create the ember model for system notifications. These notifications belong to a "Gebruiker"

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/137

### DIFF
--- a/app/models/system-notification.js
+++ b/app/models/system-notification.js
@@ -12,7 +12,7 @@ export default class SystemNotificationModel extends Model {
 
   @belongsTo('gebruiker', {
     async: true,
-    inverse: 'system-notification',
+    inverse: 'systemNotifications',
   })
   gebruiker;
 }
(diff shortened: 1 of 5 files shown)


Rahien (approved):

app/models/system-notification.js
```diff
@@ -0,0 +1,18 @@
...
+  @attr('string') subject;
+  @attr('string') message;
+
+  @attr('datetime') createdAt;
+  @attr('datetime') readAt;
+  @attr('datetime') archivedAt;
+
+  @belongsTo('gebruiker', {
+    async: true,
+    inverse: 'system-notification',
```
bfidlers, line 15: I think this should be:
`inverse: 'systemNotifications',` 
no?

JonasVanHoof: If you look at other models indeed this should be `systemNotifications` instead of `system-notifications. Changed, it looks like it didn't make a difference but now its in line with the rest :D

PR #208, opened 2024-05-10
Ben/lmb 423 bestuursorgaan detail page

bfidlers:
## Description

Updated the designs for the organen/orgaan routes. Mostly design updates, but also remove the bestuursperiod selector here, made the bewerk a modal and make it possible to edit mandaten in the table.

Screenshot 2024-05-10 at 13 17 07
Screenshot 2024-05-10 at 13 17 26
Screenshot 2024-05-10 at 13 17 35

## How to test

Check the index page of a specific bestuursorgaan, best to check both an active, inactive and a decretaal one.


JonasVanHoof:
Would remove the `bewerk` buttons as it only updates the `houders` and use that clean component you made to edit an item in draft mandataris list
image

app/controllers/organen/orgaan/index.js
```diff
@@ -1,15 +1,21 @@
...
+  toggleModal() {
+    this.isModalActive = !this.isModalActive;
+  }
 
   @action
-  selectPeriod(period) {
-    this.bestuursperiode = period.id;
+  onSave() {
+    this.isModalActive = !this.isModalActive;
+    setTimeout(() => this.router.refresh(), 1000);
```
JonasVanHoof, line 18: Hm why are we doing this ?

app/routes/organen/orgaan/index.js
```diff
@@ -1,24 +1,33 @@
...
       this.computeBestuursfuncties(mandaten),
     ]);
 
-    return {
+    const bestuursorgaanFormDefinition = getFormFrom(
+      this.store,
+      BESTUURSORGAAN_FORM_ID
+    );
+
+    return RSVP.hash({
```
JonasVanHoof, line 24: Do we need the RSVP here? All in information is awaited above? Or im still missing something about the RSVP ;p

JonasVanHoof (approved):
Looks good 👌🏻

app/controllers/organen/orgaan/index.js
bfidlers: That is an issue with caching, since we use the form-content service and not the resources directly, the frontend does not know when our resource has actually been updated so it can't refresh at the appropriate time. That's why we use an arbitrary timeout, we should fix this at some point, but I think it was difficult to do.

app/routes/organen/orgaan/index.js
bfidlers: In general if you return multiple async resources it is best to use the RSVP hash, is also something that the ember guides suggest. In this case I added this, because the formDefinition was still undefined when I tried to access it in the template (I see there is no await there, could be that just adding the await before the getFormFrom also solves it, but don't necessarily think that is better, since you don't need this for fetching of other resources)

PR #209, opened 2024-05-10
LMB- 296: Mandataris has a link to besluit

JonasVanHoof:
## Description

When the besluit is not added as linked data we want the user to set a url to the besluit

## How to test

1. Go to a person and click the detail of a mandataris
2. Correct the mandataris and add the link to the besluit
3. The link  to besluit will now be visible in the information card

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/140

## Attachments

image

image

### DIFF
--- a/app/components/mandatarissen/mandataris-card.hbs
+++ b/app/components/mandatarissen/mandataris-card.hbs
@@ -57,6 +57,13 @@
             this.formattedBeleidsdomein
           }}</div>
       </div>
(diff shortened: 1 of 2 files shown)


bfidlers (approved):
Looks good. 
We might need to find another place where we also can add this link, without needing to "correct mistakes", but that is an issue for later.

PR #210, opened 2024-05-10
Mayor selector

ruben-bottu:
## Description

We want to be able to select a mayor in the "Voorbereiding legislatuur"

## How to test

- If there is no mayor selected yet, try selecting a person that should be mayor
- If there already is a mayor selected, try changing the mayor with the person selector
- If there are multiple mayors for this "bestuursperiode", then an error message should be shown (more mayors can be created through the bestuursorgaan route)

image


app/components/person/selector.js
```diff
@@ -16,16 +16,16 @@ export default class PersonSelectorComponent extends Component {
   @tracked creatingPerson = false;
   @tracked createPersonFormDefinition;
 
-  constructor() {
```
bfidlers, line 19: Why did you remove this? This is still needed in other places where the selector is used, otherwise you never see the selected person when opening this component.
If this part is problematic for you, you should create another component, or add an option that circumvents the constructor if needed.
Screenshot 2024-05-13 at 08 24 58

app/utils/get-mandataris-status.js
bfidlers, line 1: I think this is really confusing, you rename the uris to indicate whether it is the publication status or the normal status of the mandataris, but then you combine them in one file where you don't make that distinction.

app/components/mandaat/burgemeester-selector.js
```diff
@@ -0,0 +1,109 @@
...
+    return this.bestuursorgaanInTijd.bindingEinde;
+  }
+
+  formatToDateString(dateTime) {
+    return dateTime ? moment(dateTime).format('YYYY-MM-DD') : undefined;
+  }
+
+  async getBurgemeesterMandaat() {
+    return await queryRecord(this.store, 'mandaat', {
+      filter: {
```
bfidlers, line 42: This seems a bit of overkill, maybe best to simplify this after the routing of this page is improved.

app/components/mandaat/burgemeester-selector.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  async load() {
+    const burgemeesterMandaat = await this.getBurgemeesterMandaat();
+    this.persoon = await this.loadBurgemeesterPersoon(burgemeesterMandaat);
+    console.log(this.persoon);
+  }
+
+  constructor() {
```
bfidlers, line 98: I think the constructor should be the first method in a class, this seems like it is hidden ...

bfidlers (changes requested):
Has some issues.
Also get the following error, whenever I load the prepare page. 
Screenshot 2024-05-13 at 08 24 47

And the following error when I select a person.
Screenshot 2024-05-13 at 08 27 51

Also the linter fails.

app/components/mandaat/burgemeester-selector.hbs
```diff
@@ -0,0 +1,29 @@
+{{#if this.hasMultipleBurgemeesters}}
+  <div
+    class="au-o-box no-pagination au-u-padding-bottom-none au-u-margin-bottom-none"
```
bfidlers, line 3: This no-pagination shouldn't be here, this is css for tables, don't think this has any effect here.

app/templates/verkiezingen/verkiezingsuitslag/prepare.hbs
```diff
@@ -40,13 +40,13 @@
     @showStartEndDate={{true}}
   />
 
-  <AuContent @skin="small">
+  <div class="au-o-box no-pagination au-u-padding-top-none">
```
bfidlers, line 43: This no-pagination shouldn't be here, this is css for tables, don't think this has any effect here.

bfidlers (commented):
I think the message when multiple burgemeesters are found is a bit confusing. First it is above the section of burgemeesters shouldn't it be inside this section? Also in that case no burgemeester is shown, it would probably not be that straightforward to show both, but I would at least show one.
Screenshot 2024-05-13 at 12 49 55

bfidlers (approved):
Looks good

PR #213, opened 2024-05-13
Ben/lmb 433 installatievergadering routing

bfidlers:
## Description

Updated the routing for the verkiezingen module, so this is a bit more clean. Added some helper functions to fetch bestuursorganen and updated the installatievergadering so there are multiple tables there for multiple bestuursorganen.
Screenshot 2024-05-15 at 09 18 04
Screenshot 2024-05-15 at 09 18 39

## How to test

!First update the backend and be sure the migrations have run.

Check out the verkiezingen route, see if everything is clear, if you find all relevant bestuursorganen in the installatievergadering (there should be six with burgemeester included) and test if everything still works.

## Remarks
There are still a few things that are a bit funky and should be fixed (there always is a burgemeester, whether you delete it or not, sorting is a bit broken and sorts all tables at once, the different tables should show different columns and tables should be autopopulated)

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/142


app/templates/verkiezingen/installatievergadering.hbs
```diff
@@ -0,0 +1,74 @@
...
+        <AuHeading @skin="2">Voorbereiding legislatuur</AuHeading>
+      </div>
+    </Group>
+    <Group>
+      <Mandatenbeheer::BestuursperiodenSelector
+        @options={{@model.bestuursPeriods}}
+        @selected={{@model.selectedPeriod}}
+        @onSelect={{this.selectPeriod}}
+        @prefix="Bestuursperiode"
+      />
```
JonasVanHoof, line 39: Current period is 2025 -heden is ok but when switching to a period in the past and wanting to go back to heden it is not possible

image

app/components/verkiezingen/prepare-legislatuur-section.hbs
```diff
@@ -0,0 +1,54 @@
...
+    as |Group|
+  >
+    <Group class="au-u-margin-bottom-none">
+      <div class="au-o-flow">
+        <AuHeading @skin="3">Samenstelling
+          {{@bestuursorgaan.isTijdsspecialisatieVan.naam}}</AuHeading>
+      </div>
+    </Group>
+  </AuToolbar>
+  <Verkiezingen::DraftMandatarisList
```
JonasVanHoof, line 17: The sorting is indeed a little funcky ;p the calls are happening but in the UI i don't see anything happing

app/components/verkiezingen/draft-mandataris-list.hbs
```diff
@@ -1,137 +1,139 @@
...
-          @label="Voornaam"
-        />
-        <AuDataTableThSortable
-          @field="isBestuurlijkeAliasVan.achternaam"
-          @currentSorting={{@sort}}
-          @label="Familienaam"
-        />
-        {{#if @showFractie}}
+<div {{did-update this.onInit @sort}}>
+  <div class="au-o-box no-pagination">
```
JonasVanHoof, line 2: Nice din't know that this was a thing `no-paghination`

app/components/verkiezingen/draft-mandataris-list.hbs
```diff
@@ -1,137 +1,139 @@
...
+            @label="Voornaam"
           />
-        {{/if}}
-        {{#if @showFunctie}}
           <AuDataTableThSortable
-            @field="bekleedt.bestuursfunctie.label"
+            @field="isBestuurlijkeAliasVan.achternaam"
             @currentSorting={{@sort}}
-            @label="Mandaat"
+            @label="Familienaam"
```
JonasVanHoof, line 22: Should we start with translation files so we have the same column name for every table when we want to display the firstname?  This is ok but just a question 😁

app/routes/verkiezingen.js
```diff
@@ -0,0 +1,30 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
```
JonasVanHoof, line 2: remove `inject as`

app/services/bestuursorganen.js
```diff
@@ -0,0 +1,62 @@
...
+    return await this.store.query('bestuursorgaan', queryOptions);
+  }
+
+  async getAllPoliticalBestuursorganen() {
+    return await this.store.query('bestuursorgaan', {
+      'filter[:has-no:deactivated-at]': true,
+      'filter[:has-no:is-tijdsspecialisatie-van]': true,
+      'filter[classificatie][id]':
+        this.decretaleOrganen.classificatieIds.join(','), // only organs with a political mandate
+      include: 'classificatie,heeft-tijdsspecialisaties',
```
JonasVanHoof, line 47: Im also a fan of using .join(',') here for the includes. It gives a better overview IMO

app/components/verkiezingen/prepare-legislatuur-section.hbs
```diff
@@ -0,0 +1,54 @@
...
+    @showBevoegdheid={{true}}
+    @showStartEndDate={{true}}
+  />
+
+  <AuContent @skin="small">
+    <AuButton
+      @skin="naked"
+      @icon="add"
+      {{on "click" this.createMandataris}}
+    >Toevoegen</AuButton>
```
JonasVanHoof, line 35: This is not for this PR but should we find a way that we can add the column in the component itself and only refresh the component?

JonasVanHoof (commented):
The page is looking nice. Added some remarks here and there but nothing crazy. Won't approve it as it feels as it does more than im aware of :)

app/components/verkiezingen/draft-mandataris-list.hbs
Rahien: no we won't do translation files

app/components/verkiezingen/prepare-legislatuur-section.hbs
Rahien: sorting can wait for a separate pr, we need this first

app/components/verkiezingen/prepare-legislatuur-section.hbs
Rahien: baby steps...

app/routes/verkiezingen.js
Rahien: feels like a thing we do in general and should fix in one pass. won't be stopping this pr for it

app/services/bestuursorganen.js
Rahien: there is no reason to do this though, we're not getting any additional safety on the relation names or anything so directly using a string is just as fine imo

app/templates/verkiezingen/installatievergadering.hbs
Rahien: this was an issue with the sorting: if an expected organ type is missing, the list of organs contains a null value. fixed that.

Rahien (approved):

PR #214, opened 2024-05-14
hide inappropriate fields in forms too, deal with titelvoerend in the same way as aangewezen

Rahien:
## Description

hide inappropriate fields in forms too, deal with titelvoerend in the same way as aangewezen

## How to test

open a gemeenteraadslid, it should not show rangorde or beleidsdomeinen and the status values don't include the burgemeester special cases.

image

open a schepen, it should show rangorde and beleidsdomeinen but the status should not include the burgemeester specific info. Rangorde shows a warning when incorrect values are entered.

image

open a mayor, it should not show rangorde but should show beleidsdomeinen. The status should contain all values

image

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/180


JonasVanHoof (approved):
open a gemeenteraadslid, it should not show rangorde or beleidsdomeinen and the status values don't include the burgemeester special cases. OK

PR #215, opened 2024-05-14
LMB-425: Redesign general mandataris page

JonasVanHoof:
## Description

We have new designs for the general mandataris page. (/mandatarissen/search) Add the side bar with the filters and update accordingly and come as close to the design as possible.

## How to test

1. Go to /mandatarissen/search
2. Click through on person
3. Click through mandaten
4. Filter on the available filters

## Attachments

image


ruben-bottu (commented):
Overall looks good, I am only getting some strange results with the fractie filter.
When I select some fracties I still get the mandatarissen without a fractie, and with other fractie filters I don't. Also when I select multiple fracties, I suddenly get more results.

Rahien (approved):

PR #217, opened 2024-05-15
Fold Fracties

ruben-bottu:
## Description

A mandataris can have changed fractions. If so, whenever we show a combined view of a mandataris, we should show all fractions like this: `Current (onafhankelijk)`. It was already available in the organen mandatarissen table, but it should be applied more broadly across the whole app.

## How to test

Go to a organen mandatarissen table and have a look at the fracties if they have the proper format. (Will also be applied to the mandatarissen search page in the future)

image

### DIFF
--- a/app/routes/organen/orgaan/mandatarissen.js
+++ b/app/routes/organen/orgaan/mandatarissen.js
@@ -1,5 +1,6 @@
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
+import { toUserReadableListing } from 'frontend-lmb/utils/to-user-readable-listing';
 import moment from 'moment';
 
 export default class OrganenMandatarissenRoute extends Route {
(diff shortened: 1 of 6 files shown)


app/routes/organen/orgaan/mandatarissen.js
```diff
@@ -88,80 +89,105 @@ export default class OrganenMandatarissenRoute extends Route {
...
+      foldedEnd: mandataris.einde,
+      mandataris,
+      fractie,
+      foldedFracties: fracties,
+    };
+  }
+
+  fractiesToString(currentFractie, allFracties) {
+    const sortedFracties = allFracties
+      .filter((f) => f != currentFractie && f) // TODO the "&& f" is confusing
```
ruben-bottu, line 186: Inherited from routes/organen/orgaan/mandatarissen foldMandatarissen method, but not sure why the filter works this way

Rahien (approved):

PR #220, opened 2024-05-17
add other mandates when selecting mayor too

Rahien:
## Description

add other mandates when selecting mayor too. Note: this requires a page refresh to show the new mandataris instances if  no such role exists, but since we will be prefilling the mandates with empty instances, this should not be an issue.

## How to test

Change the burgemeester. See the mandataris for the voorzitter vast bureau and voorzitter bcsd update too.
When adding multiple mandataris instances with those roles only the first one will be updated and a warning will be shown.

image


app/components/mandaat/burgemeester-selector.js
```diff
@@ -79,26 +110,47 @@ export default class MandaatBurgemeesterSelectorComponent extends Component {
...
     }
+    if (voorzitterVastBureau.length === 0) {
+      targetMandatarisses.push(
+        await this.createMandataris(this.voorzitterVastBureauMandate)
+      );
+    } else if (voorzitterVastBureau.length === 1) {
+      targetMandatarisses.push(voorzitterVastBureau[0]);
+    } else {
+      await this.setMultipleBurgemeestersError(voorzitterVastBureau);
+      targetMandatarisses.push(voorzitterVastBureau[0]);
```
bfidlers, line 138: I think this is targetMandatarisses is not super user friendly. It can be that you update the voorzitter without any warning. Or it can be that you have two voorzitters one that is the same as the burgemeester and another one as well. But if the burgemeester is the second result of the query, it will update the first voorzitter and not the one that is the same as the burgemeester.

bfidlers (approved):
One small remark, but looks good and works as advertised

PR #221, opened 2024-05-17
rework person page design (and fix designs of all description label/values)

Rahien:
## Description

rework person page design (and fix designs of all description label/values)
image

image

## How to test

Open the pages from the screenshots

### DIFF
--- a/app/styles/_shame.scss
+++ b/app/styles/_shame.scss
@@ -64,15 +64,15 @@
 // Descriptions > @todo: replace with label and p
 .au-c-description-label {
   @include au-font-size($au-h6);
+  padding: 0.8rem 0;
   font-family: $au-font;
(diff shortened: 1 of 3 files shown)


app/styles/_shame.scss
```diff
@@ -64,15 +64,15 @@
 // Descriptions > @todo: replace with label and p
 .au-c-description-label {
   @include au-font-size($au-h6);
+  padding: 0.8rem 0;
```
bfidlers, line 67: I would not add this padding like this, this makes the component look a bit skewed, with more margin on the top and the bottom of the card then on the sides.
I would just remove this padding altogether, and maybe increase the padding of the grid a bit.

bfidlers (approved):
Looks good

PR #223, opened 2024-05-17
only show rangorde input if schepen mandataris in draft

Rahien:
## Description
only show rangorde input if schepen mandataris in draftonly show rangorde input if schepen mandataris in draft

## How to test

image


JonasVanHoof (approved):
Great

PR #229, opened 2024-05-22
we don't have access to the rrn scope and requesting this will trigger an exception when logging in

Rahien:
## Description

we don't have access to the rrn scope and requesting this will trigger an exception when logging in


JonasVanHoof (approved):
Can just login and get all information in the app, also tried updating the beleidsdomeinen

PR #230, opened 2024-05-22
fix for beleidsdomein needing an array but receiving a proxy

Rahien:
## Description

fix for beleidsdomein needing an array but receiving a proxy
## How to test

update mandataris state, move to verhinderd


JonasVanHoof (approved):
Works again

PR #231, opened 2024-05-22
change links to mandataris lists when linking to person or orgaan

Rahien:
## Description

change links to mandataris lists when linking to person or orgaan

## How to test

navigate to orgaan from orgaan overview or from mandataris, should go to  list of mandatarisses in orgaan
navigate to person from mandate, should go to list of their mandates

breadcrumbs should still go to details because that's the parent route, tried the other way around, didn't make sense.


JonasVanHoof (approved):
Resolve conflicts and ok

PR #233, opened 2024-05-23
LMB-359: Create system notification UI

JonasVanHoof:
## Description

Add a badge that shows the amount of notifications in the header, make it clickable and let it route to the new system-notifications route.

On this page you can see all the notifications per tab read/unread/archived

## How to test

1. Create some system notifications for the current "gebruiker"
2. Mark as read and archive some notifications

```js
// Go to the ember data and select the current "gebruiker"
const n = $E.store.createRecord('system-notification', {
  subject: 'Onderwerp melding',
  message: 'Dit is een test melding',
  createdAt: new Date(),
  readAt: null,
  archivedAt: null,
  gebruiker: $E,
})
n.save()
```

### DIFF
--- a/app/components/system-notification-row.hbs
+++ b/app/components/system-notification-row.hbs
@@ -1,0 +1,35 @@
+{{#if this.isHidden}}
+  <td class="height-zero"></td>
+  <td class="height-zero"></td>
+  <td class="height-zero"></td>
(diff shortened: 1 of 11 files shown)


ruben-bottu (approved):
Still has some minor issues, created a ticket for them, but overall good

PR #234, opened 2024-05-23
BCSD Voorzitter Alert

ruben-bottu:
## Description

A warning is shown if the voorzitter van BCSD is not a member of RMW or Vast Bureau

## How to test

- Add a voorzitter to BCSD
- A closeable warning should be shown
- Add that person as a member (lid) to RMW or Vast Bureau and the warning should disappear without needing a refresh


app/components/verkiezingen/bcsd-voorzitter-alert.js
```diff
@@ -0,0 +1,110 @@
...
+    return this.args.bestuursorganenInTijd;
+  }
+
+  // Polling because otherwise changes in the installatievergadering data are not picked up
+  handleErrorMessage = task({ restartable: true }, async () => {
+    if (this.isDestroyed) {
+      return;
+    }
+
+    const voorzitter = await this.getVoorzitterBCSD();
```
Rahien, line 32: somehow the error message isn't shown if you add a new voorzitter that isn't in the required organs until you refresh.

Rahien (changes requested):

app/templates/verkiezingen/installatievergadering.hbs
```diff
@@ -63,7 +63,13 @@
         @bestuursorganen={{@model.bestuursorganenInTijd}}
         @sort={{this.sort}}
         @form={{@model.mandatarisNewForm}}
-      />
+      >{{#if (await bestuursorgaan.isBCSD)}}
+          <Verkiezingen::BcsdVoorzitterAlert
+            @bcsdBestuursorgaanInTijd={{bestuursorgaan}}
+            @bestuursorganenInTijd={{@model.bestuursorganenInTijd}}
+          />
+        {{/if}}
```
ruben-bottu, line 71: Is it still useful to have this if in the `yield`? Can't it be moved to `PrepareLegislatuurSection` component?

JonasVanHoof: Why are we not putting this component in the `PrepareLegislatuurSection` component? Everytime an update is done in the table it reloads and the alert check can be triggered. If we do this we get rid of the polling and it is checked on every update. 

Im I missing somehting do this?

bfidlers (approved):
Moved the warning, but looks good otherwise.

PR #237, opened 2024-05-28
Ben/lmb 475 expired mandates in search

bfidlers:
## Description

Bufgix show latest mandaat in mandataris/search page, even if it is no longer active.

## How to test

Go to the mandataris search page and check that there are no longer people in the list with an empty mandataris column.


JonasVanHoof (approved):
Still having rows with empty mandaten
image
 
 This because we check on bestuursperiode einde  greator or equal than. Chenged this to is greater than only

PR #238, opened 2024-05-28
remove hasActiveChildRoute

bfidlers:
## Description

Remove hasActiveChildRoute in fracties route.

## How to test

See if different pages in child route still work.


JonasVanHoof (approved):
Nothing strange happening with the fraction routes

PR #244, opened 2024-06-04
LMB-147: Restrictions to inputs

JonasVanHoof:
## Description

Not all restrictions can be set through the forms. Sometimes we are not using forms so but we still have to constraint the input.

## How to test

- aantal houders of mandaat needs to be limited 

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/157

## Attachments

image


ruben-bottu (approved):
Nice!

PR #246, opened 2024-06-04
LMB-445: Remove adding modified from frontend

JonasVanHoof:
## Description

As the modified is now added through the track-modified service in our app we do not have to add it in our frontend anymore.

## How to test

Make sure that you cannot find any records being updated/created with a modified property set.

I looked at the [PR](https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/173/files) for what changed when doing the opposite of this logic 

The only thing im noticing is that the data in the Ember tab shows modified as undefined when The data is not fetched again from the database => ⚠️? When refreshing and looking in the Ember tab the modified is updated. I think this is no issue? We can always add the default value for the modified property back and than it will aways be set on the ember site first and than updated in the backend. (Which will differ in a few seconds off ;p)


ruben-bottu (approved):
Approved!! :1st_place_medal:

PR #248, opened 2024-06-06
LMB-131: Cleanup form instance components

JonasVanHoof:
## Description

We have two form components to edit and create entities. The logic in these components are similar on a lot of places. We want to cleanup these components so they are still easy to understand but do not have so many duplicate code.

## How to test

1. Create a new entity with a form (/forms route)
2. Update an entity that is being updated with a form (e.g. correct mandataris)

### DIFF
--- a/app/components/form/new-instance.js
+++ b/app/components/form/new-instance.js
@@ -1,6 +1,5 @@
 import Component from '@glimmer/component';
 
-import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { service } from '@ember/service';
 import { guidFor } from '@ember/object/internals';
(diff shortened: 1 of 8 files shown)


app/components/form/new-instance.js
```diff
@@ -38,21 +33,12 @@ export default class NewInstanceComponent extends Component {
...
-    return this.formInfo !== null;
-  }
-
-  get isSaving() {
-    return this.save.isRunning;
+    this.setupNewForm.perform();
   }
 
   save = task({ keepLatest: true }, async () => {
     // TODO validation needs to be checked first before the form is actually saved
```
JonasVanHoof, line 40: Is this comment still accurate with a validation running on L44?

app/services/form-repository.js
```diff
@@ -0,0 +1,103 @@
...
+        errorMessage:
+          'Het formulier werd niet correct opgeslagen. Probeer het later opnieuw.',
+      };
+    }
+
+    return {
+      id: id,
+      errorMessage: null,
+    };
+  }
```
JonasVanHoof, line 77: Not happy with these objects. Make it that you only return at the end of the method

app/utils/is-valid-form.js
```diff
@@ -0,0 +1,27 @@
+import { validateForm } from '@lblod/ember-submission-form-fields';
+
+export function isValidFormFromFormInfo(formInfo) {
```
ruben-bottu, line 3: This function name is a bit confusing. Maybe something like `isValidFormInfo` ?

app/services/form-repository.js
```diff
@@ -0,0 +1,91 @@
...
+        }),
+      }
+    );
+
+    await timeout(RESOURCE_CACHE_TIMEOUT);
+
+    return await this._handleUpdateFetchhResult(result);
+  }
+
+  async _handleCreateFetchhResult(result) {
```
ruben-bottu, line 54: Typos! `_handleCreateFetchhResult` and `_handleUpdateFetchhResult` have too many h'es
besides that, I would rename it to `handleCreateResponse` and `handleUpdateResponse`

ruben-bottu (approved):
Left a few comments, but overall looks good

app/utils/is-valid-form.js
JonasVanHoof: Renamed to `isValidForm`

app/services/form-repository.js
JonasVanHoof: Is it validating something? Ill change it to `_handleCreateResult` and `_handleUpdateResult`

PR #249, opened 2024-06-10
LMB-133: Skeleton loaders

JonasVanHoof:
## Description

Instead of showing the three dots or loading text when we are waiting for our data to show up use skeletons so the user sees whats coming.

## How to test

1. Click through the app and see that everything works as expected + when you are waiting on data that it shows you a skeleton for the components to load.
2. You can also search on "loading" in the frontend project to make sure I didn't miss any.
3. Or search for "AuLoader" (here I let the auth frontend is what it is)

### DIFF
--- a/app/components/bestuursperiode-new-button.hbs
+++ b/app/components/bestuursperiode-new-button.hbs
@@ -18,5 +18,5 @@
     </div>
   </AuModal>
 {{else}}
-  Loading...
+  <Skeleton::Forms::BestuursperiodeNew />
 {{/if}}
--- a/app/components/bestuursperiode-new-button.hbs
+++ b/app/components/bestuursperiode-new-button.hbs
@@ -2,6 +2,7 @@
   <AuButton {{on "click" this.openModal}}>
     Voeg bestuursperiode toe
   </AuButton>
+  <AuModalContainer />
   <AuModal
     @title={{this.title}}
     @modalOpen={{this.isModalOpen}}
--- a/app/components/form/cancel-with-confirm.hbs
+++ b/app/components/form/cancel-with-confirm.hbs
@@ -1,5 +1,6 @@
 <AuButton @skin="secondary" {{on "click" this.cancel}}>{{yield}}</AuButton>
 
+<AuModalContainer />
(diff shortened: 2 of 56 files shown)


app/components/bestuursperiode-new-button.hbs
```diff
@@ -2,6 +2,7 @@
   <AuButton {{on "click" this.openModal}}>
```
ruben-bottu, line 2: Is it possible that this component is not used anymore?

app/components/form/cancel-with-confirm.hbs
```diff
@@ -1,5 +1,6 @@
 <AuButton @skin="secondary" {{on "click" this.cancel}}>{{yield}}</AuButton>
 
+<AuModalContainer />
```
ruben-bottu, line 3: Nice catch! Didn't know this was necessary. I think we mostly use the one in application.hbs

app/components/form/instance.js
```diff
@@ -245,5 +246,8 @@ export default class InstanceComponent extends Component {
     };
     formStore.registerObserver(onFormUpdate);
     onFormUpdate();
+    await timeout(55);
```
ruben-bottu, line 249: Seems like a magic number

ruben-bottu (approved):
Seems to work as expected. Haven't been able to see/find all skeleton components, but the ones I saw looked fine. Added some small comments.

app/components/form/cancel-with-confirm.hbs
JonasVanHoof: It was very annoying adding these everywhere as the frontend keeps failing on it when it is not there 😓

app/components/bestuursperiode-new-button.hbs
JonasVanHoof: I can't seem to find an implementation. Will remove it in a separate PR

app/components/form/instance.js
JonasVanHoof: Removed it! I added an await to this.registerObserver and does the same trick

PR #252, opened 2024-06-14
LMB-555 - legislatuur disables mandatees

JonasVanHoof:
## Description

Right now the wrong resources are locked when dealing with the preparation of a legislature.

It should be the case that mandatarissen (and also the fracties) are locked whenever they have a corresponding preparation of a legislature that is still in preparation or ready for review.

So when you are preparing a legislature, the mandatarissen you add in this page should not be editable in  their corresponding mandataris page. When this preparation is done and handled, they should become editable.

For mandatarissen that have no corresponding legislature, they should still be editable.

## How to test

When a legislatuur is not behandeld the buttons should be disabled for that besuursperiode

### DIFF
--- a/app/controllers/mandatarissen/persoon/mandataris.js
+++ b/app/controllers/mandatarissen/persoon/mandataris.js
@@ -17,7 +17,8 @@ export default class MandatarissenPersoonMandatarisController extends Controller
 
   @tracked isChanging;
   @tracked isCorrecting;
-  @tracked isInTeBehandelenLegislatuur;
+  @tracked periodeHasLegislatuur;
+  @tracked behandeldeVergaderingen;
 
   get bestuursorganenTitle() {
     const bestuursfunctie = this.model.mandataris.bekleedt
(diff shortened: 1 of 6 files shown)


bfidlers (commented):
It is an improvement, but you miss the case where there is no installatievergadering. If you have a bestuursperiode that has no installatievergadering, the isInTeBehandelenLegislatuur you use will result in true. So mandatarissen won't be editable, but in that case they should be, otherwise they are never editable, since they have no installatievergadering.
If it makes it easier, it is okay for me if you just check if the installatievergadering is in ' "te behandelen" state, and count the "klaar voor behandeling" as "behandeld" that should make it possible to keep your logic but just use a different filter and invert the length === 0 check.
But if you find another solution you like more, go ahead.

JonasVanHoof:
> It is an improvement, but you miss the case where there is no installatievergadering. If you have a bestuursperiode that has no installatievergadering, the isInTeBehandelenLegislatuur you use will result in true. So mandatarissen won't be editable, but in that case they should be, otherwise they are never editable, since they have no installatievergadering. If it makes it easier, it is okay for me if you just check if the installatievergadering is in ' "te behandelen" state, and count the "klaar voor behandeling" as "behandeld" that should make it possible to keep your logic but just use a different filter and invert the length === 0 check. But if you find another solution you like more, go ahead.

Good catch didn't think about that one.. I updated the code to first check if there is a legislatuur for the current (selected) bestuursperiode. If that is not the case it will not disable any buttons. If it is, it will go and check if there are "behandelde vergaderingen". When there are no "behandelde vergaderingen" the buttons need to be disabled because than it is still "in behandeling" of "klaar voor vergadering".

bfidlers (approved):
Looks good

PR #253, opened 2024-06-14
LMB 491 | Mandataris without person breaks burgemeester selector

JonasVanHoof:
## Description

When no burgemeester is selected it errors in the console as it cannot fetch its data. 

## How to test

1. Do not select a person in the legislatuur page; it should not throw an error
2. Start selecting a person but than cancel the model; it should not throw an error

## Attachments

**Error**
image

### DIFF
--- a/app/components/mandaat/burgemeester-selector.js
+++ b/app/components/mandaat/burgemeester-selector.js
@@ -42,7 +42,7 @@ export default class MandaatBurgemeesterSelectorComponent extends Component {
   async load() {
     await this.loadBurgemeesterMandates();
 
-    if ((this.burgemeesterMandate && this, this.voorzitterVastBureauMandate)) {
+    if (this.burgemeesterMandate && this.voorzitterVastBureauMandate) {
       this.persoon = await this.loadBurgemeesterPersoon();
     }
   }
--- a/app/components/mandaat/burgemeester-selector.js
+++ b/app/components/mandaat/burgemeester-selector.js
@@ -165,11 +165,11 @@ export default class MandaatBurgemeesterSelectorComponent extends Component {
 
   @action
   async onUpdate(persoon) {
-    if (!persoon) {
+    this.persoon = persoon;
+
+    if (!this.persoon || !this.targetMandatarisses) {
       return;
     }
-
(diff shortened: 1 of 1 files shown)


app/components/mandaat/burgemeester-selector.js
```diff
@@ -67,15 +71,35 @@ export default class MandaatBurgemeesterSelectorComponent extends Component {
...
-        BESTUURSFUNCTIE_VOORZITTER_VAST_BUREAU_ID
+
+    if (!burgemeester) {
+      this.toaster.notify(
+        `Geen ${mapping[bestuursfunctieId] ?? ''} mandaat gevonden.`,
+        'Opgelet!',
+        { type: 'warning', timeOut: 5000 }
       );
-    });
+    }
```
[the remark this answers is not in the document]
JonasVanHoof: We do not want this

app/components/mandaat/burgemeester-selector.js
```diff
@@ -41,7 +41,10 @@ export default class MandaatBurgemeesterSelectorComponent extends Component {
 
   async load() {
     await this.loadBurgemeesterMandates();
-    this.persoon = await this.loadBurgemeesterPersoon();
+
+    if ((this.burgemeesterMandate && this, this.voorzitterVastBureauMandate)) {
```
bfidlers, line 45: But weird syntax, I would just use 
`if (this.burgemeesterMandate && this.voorzitterVastBureauMandate) {`
if that is possible

bfidlers (approved):
Couldn't reliably reproduce the error, but can't reproduce it with this fix, so seems good.
One small code style remark however.

app/components/mandaat/burgemeester-selector.js
JonasVanHoof: This is going to be added when selecting a suggestion I think, I was lucky it didn't break it ;p

JonasVanHoof:
Keep in mind that when selecting a persoon and the correct mandates are not available the user will still see "Selecteer een persoon" again as they could not set it to the mandatees. Maybe add a warning for this in the future.

PR #254, opened 2024-06-14
LMB-377 require mandaat access roles

bfidlers:
## Description

Require mandaat access roles for all different modules in our application.

## How to test

You can test if you can still access all cards in the index page. Mocking not having roles is not really worth it.


JonasVanHoof (changes requested):
Looks clean. Limiting the routes is nice but I assume that we do not show the modules when they do not have the permission is also something todo in this PR?

JonasVanHoof:
Created a ticket for disabling the cards in Jira

PR #256, opened 2024-06-26
move creation beeindigd mandataris status code to migration

bfidlers:
## Description

The beeindigd mandataris status code was moved to a migration, this has some influence on the mandataris status service in the frontend.

## How to test

See if the update state in mandataris still works (especially the beeindigd state), or if you can select the beeindigd state is available in a mandataris form.

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/165


ruben-bottu:
We might want to make it a constraint that the end date is required when creating a Beëndigd mandataris?

PR #257, opened 2024-06-26
Ben/lmb 531 mandataris search

bfidlers:
## Description

The table in the mandataris search page is now fetched based on mandatarissen instead of on people. This makes the filters more accurate and useful. Also added an active mandataris filter. 

## How to test

Go to the mandataris search page and see if the filters work and you get the same people as before (might be less, because filters now work correctly) but without filters results should stay the same.

### DIFF
--- a/app/components/mandatarissen/persoon-table.hbs
+++ b/app/components/mandatarissen/persoon-table.hbs
@@ -9,12 +9,12 @@
   <t.content as |c|>
     <c.header>
       <AuDataTableThSortable
-        @field="gebruikteVoornaam"
+        @field="is-bestuurlijke-alias-van.gebruikteVoornaam"
         @currentSorting={{@sort}}
         @label="Naam"
       />
       <AuDataTableThSortable
-        @field="achternaam"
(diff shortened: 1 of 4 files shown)


ruben-bottu:
I don't see the number of entries anymore at the bottom of the table, e.g. `1 - 15 van 15`

ruben-bottu:
The filters work much better now, the fractie selector and "Herstel alle filters" are not broken anymore.

ruben-bottu (changes requested):
Missing number of entries at the bottom of the table, e.g. `1 - 15 van 15`

ruben-bottu (approved):

bfidlers:
> I don't see the number of entries anymore at the bottom of the table, e.g. `1 - 15 van 15`

This is because we fetch mandatarissen instead of people, but show people. This means all instances are fetched in one go, which gives us no pagination. Which makes the pagination footer less usefull. It is possible to add the nb of people here, but we won't have the option to go to a next page or something like that.

PR #258, opened 2024-06-27
Ben/lmg 477 mandataris search installatievergadering

bfidlers:
## Description

Mandataris search page is not available if the legislature is still in preperation. This is implemented by disabling the disabling the bestuursperiode in the selector.

Screenshot 2024-06-27 at 12 54 34

## How to test

Test for a bestuurseenheid with installatievergadering if the bestuursperiode in the mandataris search route is disabled. And if it becomes enabled again if you update the state of the prepare legislature to behandeld.

### DIFF
--- a/app/controllers/mandatarissen/search.js
+++ b/app/controllers/mandatarissen/search.js
@@ -41,8 +41,8 @@ export default class MandatarissenSearchController extends Controller {
   }
 
   @action
-  selectPeriod(period) {
-    this.bestuursperiode = period.period.id;
+  selectPeriod(periodMap) {
+    this.bestuursperiode = periodMap.period.id;
   }
 
   @action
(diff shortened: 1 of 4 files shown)


app/controllers/mandatarissen/search.js
```diff
@@ -42,7 +42,7 @@ export default class MandatarissenSearchController extends Controller {
 
   @action
   selectPeriod(period) {
-    this.bestuursperiode = period.id;
+    this.bestuursperiode = period.period.id;
   }
```
JonasVanHoof, line 46: Little strange that it is periode.period? Maybe as you did below also change it to periodMap? periodMap.period.id

app/routes/mandatarissen/search.js
```diff
@@ -25,6 +26,22 @@ export default class MandatarissenSearchRoute extends Route {
       params.bestuursperiode
     );
 
+    const periodMap = await Promise.all(
```
JonasVanHoof, line 29: Do we need to add a comment that this is done for the power-select component?

JonasVanHoof (approved):
Looks good 🚀

PR #263, opened 2024-07-03
Ben/lmb 512 titelvoerend vs verhinderd status

bfidlers:
## Description

The mandataris status verhinderd is only available for not burgemeesters and the status titelvoerend is only available for burgemeesters.

## How to test

Every place a mandataris status code selector is used, it should only be possible to select a status that is valid with the selected mandate. This can be done in the form for creating a new mandataris or in the form for updating the status of a mandataris. 
Also if you update the status of a burgemeester to titelvoerend, this should also activate the possibility to add a replacement.


JonasVanHoof (approved):
OKidoki ✅

Should we create a migration to change the verhinderd statusses of a Burgemeesters to be titelvoerend? 
image

bfidlers:
> OKidoki ✅
> 
> Should we create a migration to change the verhinderd statusses of a Burgemeesters to be titelvoerend? image

Good catch, I'll reopen the story and do that.

PR #264, opened 2024-07-04
LMB-588 | Redesign legislatuur page

JonasVanHoof:
## Description

The legislatuur got a new design and we need to implement it.

## How to test

Look at the descirption in the ticket and the figma design to make sure I did not miss anything.


JonasVanHoof:
The beleidsdomeinen are not editable in the mandataris new form. cc: @bfidlers

app/components/verkiezingen/draft-mandataris-list.js
```diff
@@ -62,52 +63,6 @@ export default class DraftMandatarisListComponent extends Component {
     this.mandatarissen = this.store.query('mandataris', queryParams);
```
JonasVanHoof, line 63: Just saw this because it throw an error  on my logic but there is an await missing.. Updated this on my new branch. Did you have any issues with this? cc: @bfidlers

bfidlers (approved):
Looks good.

PR #266, opened 2024-07-10
LMB-481 | Update UI to show besluit

JonasVanHoof:
## Description

We want to make the link to the besluit available for the user when the status of the mandataris is bekrachtigd.

Also removed the info icon next to "contacteer technische dienst" as UI wise this was very ugly

## How to test

1. Add a link to bewijs by filling a url in into the link to besluit field when correcting mistakes.
2. Now the "Bekijk bewijs" link is shown next to the status pill and goes to that link in a new window

## Attachments

image


bfidlers (approved):
Looks good

PR #268, opened 2024-07-12
LMB-465 | Remove sorting on tables in legislatuur

JonasVanHoof:
## Description

We do not want sorting on the tables in the legislatuur page. When a rangorde is shown in the tables the values should be sorted by that property. This is logic that already existed so we just need to remove the sorting in the columns and the parameter in the query. NBot that the burgmeester selector is also shown as a table from now on so the sorting should also be removed there. (This should always be one value so it's in general a little stupid to be able to sort this ;p)

## How to test

Go to the legislatuur page and see that you cannot sort the values anymore

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/267 => **Created from this PR**

### DIFF
--- a/app/components/verkiezingen/draft-mandataris-list.js
+++ b/app/components/verkiezingen/draft-mandataris-list.js
@@ -22,20 +22,16 @@ export default class DraftMandatarisListComponent extends Component {
   }
 
   get resortedMandatarissen() {
-    if (!this.args.sort || this.args.sort.indexOf('rangorde') < 0) {
-      return this.mandatarissen;
+    if (!this.mandatarissen) {
+      return [];
     }
-    const sorted = orderMandatarissenByRangorde([...this.mandatarissen]);
-    if (this.args.sort.startsWith('-')) {
-      return sorted.reverse();
-    }
-    return sorted;
+
+    return orderMandatarissenByRangorde([...this.mandatarissen]);
   }
 
   @action
   async onInit() {
     const queryParams = {
-      sort: this.args.sort,
       page: {
         number: 0,
         size: 9999,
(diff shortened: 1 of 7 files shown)


app/components/verkiezingen/draft-mandataris-list.js
```diff
@@ -22,20 +22,16 @@ export default class DraftMandatarisListComponent extends Component {
   }
 
   get resortedMandatarissen() {
-    if (!this.args.sort || this.args.sort.indexOf('rangorde') < 0) {
-      return this.mandatarissen;
+    if (!this.mandatarissen) {
```
JonasVanHoof, line 25: Could also give it a start value when defining the tracked property but i think its better to just check it here if it is set so we do not expect default behaviour of variables.

app/components/verkiezingen/draft-mandataris-list.js
```diff
@@ -60,7 +56,7 @@ export default class DraftMandatarisListComponent extends Component {
       queryParams['filter']['is-bestuurlijke-alias-van'] = this.args.filter;
     }
 
-    this.mandatarissen = this.store.query('mandataris', queryParams);
+    this.mandatarissen = await this.store.query('mandataris', queryParams);
```
JonasVanHoof, line 59: We should await this haha else our sortedMandataris getter on L24 is throwiung an error that a promise is not iterable ;p

bfidlers (approved):
I'm not sure if I like the styling of these plain headers. The sortable ones were a bit more spacious etc. I liked this, but this PR is certainly okay, does what it says it should.

PR #271, opened 2024-07-12
LMB-591 | Update replacement status of burgemeester

JonasVanHoof:
## Description

When a burgemeester of aangewezen burgmeester gets titelvoerend the mandataris status of the replacement should be set to waarnemend.

## How to test

Open a burgemeester and set to titelvoerend status. Select a replacer and the mandataris of the replacer should be a waarnemend status.


app/services/mandataris.js
```diff
@@ -42,12 +44,26 @@ export default class MandatarisService extends Service {
...
   ) {
+    let mandatarisStatus = toReplace.status;
+
+    const mandaat = await toReplace.bekleedt;
     const existing = await this.getOverlappingMandate(
       toReplace,
       replacementPerson
     );
+
+    if (mandaat && mandaat.isBurgemeester) {
```
bfidlers, line 55: I think the mandaat check is a bit overkill, but can't hurt I guess!

app/services/mandataris.js
```diff
@@ -42,12 +44,26 @@ export default class MandatarisService extends Service {
...
+      mandatarisStatus = this.store.findRecord(
+        'mandataris-status-code',
+        MANDATARIS_WAARNEMEND_STATE_ID
+      );
+    }
+
     if (existing) {
-      return existing;
+      const mandataris = await this.store.findRecord('mandataris', existing.id);
+      mandataris.status = mandatarisStatus;
```
bfidlers, line 64: you can just use existing.status here, no?

bfidlers (changes requested):
you do two calls to get the same mandataris I think, would delete the second call.
Apart from that, looks great!

app/services/mandataris.js
JonasVanHoof: Hmm I'll try again but that didn't do it before

JonasVanHoof: It didn't update it if i do this
image
```js
    if (existing) {
      // const mandataris = await this.store.findRecord('mandataris', existing.id);
      existing.status = mandatarisStatus;
      existing.save();
      console.log('existing', existing);
      console.log('existing', await existing.status);
      return existing;
    }

bfidlers: hmmm that is weird
didn't run it yet no

bfidlers: Figured it out, you can set it directly on the exisiting mandataris, but to do that, you need to await your findRecord on line 56. Not sure why this results in different behaviour with existing and the new fetch ... Maybe because you need to wait anyway if you fetch a new record?

JonasVanHoof: Jup works indeed ! Not sure why I didn't add the await before my store.findRecord :/

bfidlers (approved):
Great! Didn't test the last changes, but those are the same as what I ran locally, so should be fine.

PR #272, opened 2024-07-15
LMB-622 | When no mandatarissen in source table show warning when syncing

JonasVanHoof:
## Description

A user is able to press the sync button on the legislatuur page. When no mandatarissen are foudn to sync nothing happens. As disabling the sync button would be alot of logic for some simple user interaction we are showing a warning toaster message when there are no mandatarissen to sync.

## How to test

Go to legislatuur page and press the sync button when there are not mandatarissen in the table.

## Attachments

image


bfidlers (approved):
Great, works as expected!

PR #273, opened 2024-07-15
Ben/lmb 595 bug fracties aanmaken

bfidlers:
## Description

It is no longer possible to add fracties if the selected bestuursperiode has no bestuursorganen. This is disabled because if this is possible these fracties are not really useful, they aren't shown, because fracties are only fetched based on bestuursorganen and these fracties would never be added to a bestuursorgaan, even if new bestuursorganen were added, so they have no point at all. 
This is disabled in two places. In the organen page: 
Screenshot 2024-07-15 at 15 16 24
And in the bestuursperiode selector of the fractie page.

## How to test

Go to the organen page and see what happens with the beheer fracties button depending on which bestuursperiode is selected. Also if you go to the beheer fracties page, the bestuursperiode selector only shows bestuursperiods for which it makes sense to edit fracties (they should have bestuursorganen and they should not have an active prepare legislature page.

### DIFF
--- a/app/routes/organen/fracties.js
+++ b/app/routes/organen/fracties.js
@@ -28,7 +28,7 @@ export default class FractiesRoute extends Route {
 
     const bestuursPeriods = await this.store.query('bestuursperiode', {
       sort: 'label',
-      'filter[:has:heeft-bestuursorganen-in-tijd]': 'true',
+      'filter[:has:heeft-bestuursorganen-in-tijd]': true,
       include: 'installatievergaderingen',
     });
 
--- a/app/routes/organen/fracties.js
+++ b/app/routes/organen/fracties.js
@@ -5,7 +5,10 @@ import { action } from '@ember/object';
 
 import { getFormFrom } from 'frontend-lmb/utils/get-form';
 import { FRACTIE_FORM_ID } from 'frontend-lmb/utils/well-known-ids';
-import { FRACTIETYPE_SAMENWERKINGSVERBAND } from 'frontend-lmb/utils/well-known-uris';
+import {
+  FRACTIETYPE_SAMENWERKINGSVERBAND,
+  INSTALLATIEVERGADERING_BEHANDELD_STATUS,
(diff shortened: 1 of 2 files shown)


app/routes/organen/fracties.js
```diff
@@ -25,9 +28,35 @@ export default class FractiesRoute extends Route {
 
     const bestuursPeriods = await this.store.query('bestuursperiode', {
       sort: 'label',
+      'filter[:has:heeft-bestuursorganen-in-tijd]': 'true',
```
JonasVanHoof, line 31: NITPICK: Does true need to be a string?

app/routes/organen/fracties.js
```diff
@@ -25,9 +28,35 @@ export default class FractiesRoute extends Route {
...
+          }
+          return {
+            value: period,
+            valid,
+          };
+        })
+      )
+    )
+      .filter((v) => v.valid)
+      .map((data) => data.value);
```
JonasVanHoof, line 56: This is great 😁

JonasVanHoof (approved):
Looks good! Tested the frontend to 👌🏻

app/routes/organen/fracties.js
bfidlers: indeed it doesn't

app/routes/organen/fracties.js
bfidlers: Improved it a little bit

JonasVanHoof: Alright, indeed a cleaner piece of code

JonasVanHoof:
I was thinking should we show a message to the user why these options are disabled? I was just thinking that we disable the fractie beheer button for two reasons now 🤔

bfidlers:
> I was thinking should we show a message to the user why these options are disabled? I was just thinking that we disable the fractie beheer button for two reasons now 🤔

Yeah might be getting a bit confusing for the user, will mention it in the standup so Tim can look into it.

PR #274, opened 2024-07-15
LMB-553 | Person, mandaat and  fractie selector support warning and error validations

JonasVanHoof:
## Description

We did not see any visual errors when a validation was added through the ttl code (form) for person, mandaat and fracties. These custoim components now display errors when not valid. 

## How to test

When the field for person, fractie or mandaat is not filled in you should not be able to save the form

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/178

## Attachments

image


JonasVanHoof:
Not sure why the pipelines are failing... it's something with the reference to the branch

bfidlers:
> Not sure why the pipelines are failing... it's something with the reference to the branch

Doesn't really look like they failed on the code, but on some connection issue. Just retriggering the pipeline often also works.

bfidlers (approved):
Nice, looks great!

PR #275, opened 2024-07-15
LMB-597 Fix mandate for mandataris not showing in mandataris search

bfidlers:
## Description

In the mandataris search page, some mandatarissen did not seem to have a visible mandate. This PR fixes this.
In the current state mandates are linked to a bestuursperiode, so it is no longer necessary to manually check the date of a mandate, which did sometimes classify a mandataris as not belonging to the correct bestuursperiode, because it had an invalid date.

## How to test

Best to first find a bestuurseenheid where mandataris search has a mandataris with an empty mandate column, then spin up this new branch and see it now does have a mandate filled in. All mandatarissen in the mandataris search page should have a mandate.


JonasVanHoof (approved):
Haha nice fix :D

Guy didn't have a mandaat in the mandataris search page and with your branch it has a mandaat set

PR #278, opened 2024-07-16
remove check if mandaat of replacement is burgemeester

bfidlers:
## Description

If you select a replacement, the status of this replacement is now always waarnemend.

## How to test

Update the state of a mandataris to verhinderd and add a replacement. Check that the status of this mandataris is correct.


JonasVanHoof (approved):
Seems to be working 👌🏻

PR #281, opened 2024-07-17
LMB-130 | Feature flags

JonasVanHoof:
## Description

Add feature flags to our app in the same way organisatie-portaal did (link in ticket). I created a test feature flag for the forms module on the overview page.

## How to test

1. Go to the overview page and normally the forms module will not be shown.
2. Add the feature flag to your url `?feature-show-forms-module-on-overview=false` http://localhost:4200/?feature-show-forms-module-on-overview=true and the forms module should be visible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Frontend of the lokaal mandatenbeheer application
 
 Feature flags are used to enable/disable features in the application. They are defined in config/environment.js.
 
-```
+```javascript
 // in config/environment.js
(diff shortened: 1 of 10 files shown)


JonasVanHoof:
Should we copy over the feature section into our Readme?

bfidlers (approved):
Works as advertised!
Updated the readme and also extended the use of the feature flag `show-forms-module` a bit.

PR #282, opened 2024-07-18
LMB-616 | Update person fractie relation when needed

JonasVanHoof:
## Description

Fractie on a person needs to be updated when a change in mandataris or fractie is done. Update it on all the possible places.

## How to test

1. Create a new person
2. Add a new mandataris to it in the current bestuursperiode (2019 - 2025)  => fractie another than onafhankelijk
3. Add a new mandataris to it in another bestuursperiode (2012 - 2019)  => fractie another than onafhankelijk
4. When going to the mandaten of that  person a blue button to become onafhankelijk should be visible 
5. When you press that button both the mandatarissen should go to onafhankelijke fractie
   5.1 The onafhankelijke fractie should have the same uri (check by changing naam to uri in history-rows.hbs)
6. When looking at the fractie value on your person it should say the same URI
7. When wijziginging of mandataris fractie  in the persons currentFractie schould be updated (current bestuursperiode)
8. When updating the fractie of  the second madnataris from another bestuursperiode and looking back at the person it should  still have the fractie URI of the madntaris in current bestuursperiode and not the URI that you just changed it to.
9. All of this done should solve the issue
 
## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/28 **mandataris service**


app/repositories/fractie.js
```diff
@@ -0,0 +1,31 @@
...
+      }
+    );
+
+    const jsonReponse = await response.json();
+
+    if (response.status !== 201) {
+      throw new Error(jsonReponse.message);
+    }
+
+    return jsonReponse.uri;
```
JonasVanHoof, line 29: IN the repositories we are always getting back the uri/id of the entity, thinking of getting the ember model in the repository aswell so we can contineu with the model. Not sure if it is worth it

app/repositories/mandataris.js
```diff
@@ -0,0 +1,34 @@
...
+    const jsonReponse = await response.json();
+
+    if (response.status !== STATUS_CODE.OK) {
+      console.error(jsonReponse.message);
+      throw jsonReponse.message;
+    }
+
+    return {
+      uri: jsonReponse.uri,
+      id: jsonReponse.id,
```
JonasVanHoof, line 31: As in this the usecase returning the model would make the mandataris service easier?

app/repositories/persoon.js
```diff
@@ -0,0 +1,34 @@
+import { API, STATUS_CODE } from 'frontend-lmb/utils/constants';
+
+export default class PersoonRepository {
+  async findOnafhankelijkeFractie(persoonId) {
+    const response = await fetch(
+      `${API.MANDATARIS_SERVICE}/personen/${persoonId}/onafhankelijke-fractie`
```
JonasVanHoof, line 6: Specifying that the used api is the mandataris service

app/services/fractie.js
```diff
@@ -1,57 +1,12 @@
...
 
 import { service } from '@ember/service';
-import { fold } from 'frontend-lmb/utils/fold-mandatarisses';
-
-import { FRACTIETYPE_ONAFHANKELIJK } from 'frontend-lmb/utils/well-known-uris';
 
 export default class FractieService extends Service {
   @service store;
 
-  async createOnafhankelijkeFractie(bestuursorganen, bestuurseenheid) {
```
JonasVanHoof, line 11: Moved to the mandataris service and in the frontend to the repositories

app/controllers/mandatarissen/persoon/mandaten.js
```diff
@@ -49,23 +57,18 @@ export default class MandatarissenPersoonMandatenController extends Controller {
...
 
-  checkFracties = task(async (foldedMandatarissen) => {
+  checkFracties = task(async (mandatarissen) => {
     this.canBecomeOnafhankelijk = false;
     this.possibelOnafhankelijkeMandatarissen = [];
-    for (const fold of foldedMandatarissen) {
-      const isActive = await this.mandatarisService.isMandatarisActive(
-        fold.mandataris
-      );
+    for (const mandataris of mandatarissen) {
```
bfidlers, line 63: Shouldn't we be able to do this as one call?

JonasVanHoof: a method `filterNonOnafhankelijkeandatarissen()` could be added to mandataris yes

app/repositories/mandataris.js
```diff
@@ -0,0 +1,34 @@
...
+
+    if (response.status !== STATUS_CODE.OK) {
+      console.error(jsonReponse.message);
+      throw jsonReponse.message;
+    }
+
+    return jsonReponse.isActive;
+  }
+
+  async getBestuursperiode(mandatarisId) {
```
bfidlers, line 18: I think it is not worth doing something like this, we only use this in the fractie repository and it gives us a headache to actually retrieve the model again, so just move this logic to the mandataris-service, when we want to update the currentFractie, we give a mandataris, which is linked to a bestuursperiode, so you don't need to pass this.

app/services/mandataris.js
```diff
@@ -143,20 +146,15 @@ export default class MandatarisService extends Service {
...
-
-    const status = await foldedMandataris.mandataris.status;
-    return status && status.uri !== MANDATARIS_BEEINDIGD_STATE;
+  async updateCurrentFractie(mandatarisId) {
+    const mandataris = await this.store.findRecord('mandataris', mandatarisId);
+    const persoon = await mandataris.isBestuurlijkeAliasVan;
+    const bestuursperiode = await this.mandatarisRepository.getBestuursperiode(
+      mandataris.id
+    );
+    await this.persoonRepository.updateCurrentFractie(
```
bfidlers, line 155: Now I get why you fetch the bestuursperiode, because it makes more sense to call the mandataris service with a person instead of a mandataris, but I still think it makes more sense to just call the mandataris service just with the mandataris. (or maybe mandataris and person if you want).

app/services/fractie.js
bfidlers, line 1: Not sure if it still makes sense that this is a service, feels more like a util to me.

app/controllers/mandatarissen/persoon/mandaten.js
bfidlers, line 1: I think more of these flows should just move to the mandatarisService as one, not all these independent calls, but that might be something to look into at a later stage.

bfidlers (approved):
Few comments here and there, looks good otherwise on this side. I'll review the mandataris-service seperately. 

We do need to be careful about different bestuursperiodes however. This is not included yet. Which might give some issues here and there.

Seems to work as expected.

app/controllers/mandatarissen/persoon/mandataris.js
```diff
@@ -61,6 +62,7 @@ export default class MandatarissenPersoonMandatarisController extends Controller
     store.parse(instanceTtl, SOURCE_GRAPH, 'text/turtle');
     const mandatarisUri = this.model.mandataris.uri;
 
+    await this.mandatarisService.updateCurrentFractie(this.model.mandataris.id);
```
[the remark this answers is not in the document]
bfidlers: no this is a different form, this is not update state.

Rahien:
also replaced by this one? https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/295

PR #286, opened 2024-07-25
Ben/lmb 548 min max nb members

bfidlers:
## Description

Extend mandate model to make it possible to add a minimum and maximum aantal houders. Also show this in the bestuursorgaan detail page.

Screenshot 2024-07-23 at 18 13 01

## How to test

Run the migration in the linked PR and check a few bestuursorganen to see if they have actually a minimum and maximum number of members. Toegevoegde schepenen will never have a min and max and only very few rmw's will have these filled in. Not all bestuursorganen will have filled in data, this is based on the following Excel, so if you don't find data for a bestuursorgaan, it is probably because it is not in this file. https://vlaamseoverheid.sharepoint.com/:x:/r/sites/ABB_LMB/_layouts/15/doc.aspx?sourcedoc=%7B5bd175d5-3dbe-4e41-9adc-51552b96b9c3%7D&action=edit

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/181

### DIFF
--- a/app/components/mandatenbeheer/bestuursorgaan-mandaten.hbs
+++ b/app/components/mandatenbeheer/bestuursorgaan-mandaten.hbs
@@ -9,8 +9,10 @@
     <t.content as |c|>
       <c.header>
         <th>Naam</th>
-        <th>Min aantal houders</th>
-        <th>Max aantal houders</th>
+        {{#if this.showMinMax}}
+          <th>Min aantal houders</th>
+          <th>Max aantal houders</th>
+        {{/if}}
         <th>Aantal houders</th>
         <th></th>
       </c.header>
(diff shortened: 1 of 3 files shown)


app/components/mandatenbeheer/bestuursorgaan-mandaten.hbs
```diff
@@ -9,14 +9,22 @@
...
+        <th>Aantal houders</th>
         <th></th>
       </c.header>
       <c.body as |mandaat|>
         <td class="au-u-visible-small-up">
           {{mandaat.bestuursfunctie.label}}
         </td>
 
+        <td class="au-u-visible-small-up">
+          {{mandaat.minAantalHouders}}
```
JonasVanHoof, line 23: Should we hide it when its not set? This is only set in bestuursperiode 2025 - heden so all the other just show an empyt filed.  We could also set Ongekend text when not specified

image

JonasVanHoof (approved):
Looks great 🚀

app/components/mandatenbeheer/bestuursorgaan-mandaten.hbs
bfidlers: good point, will do.

PR #287, opened 2024-07-26
LMB - 617 | Only show current fractie of persoon

JonasVanHoof:
## Description

When a new mandataris is created the user should only see the fractions for that user, that means that there will be two ( fractie + onafhankelijke fractie ). When we do correcting of a mandataris all fractie sin the bestuursperiode should be visible and than we change a mandataris (a new mandataris is created here) you should again only see the fracties for that person.

## How to test

1. Create a mandataris for a person
2. Update a mandataris
3. Correct a mandataris

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/282 => build on this PR

### DIFF
--- a/app/components/mandatarissen/fractie-selector.hbs
+++ b/app/components/mandatarissen/fractie-selector.hbs
@@ -11,7 +11,7 @@
   @onChange={{this.select}}
   @onClose={{@onClose}}
   @options={{this.fractieOptions}}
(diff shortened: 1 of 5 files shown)


Rahien:
I believe this PR has been replaced by this one: https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/295

PR #292, opened 2024-08-01
LMB-635 only show active mandatarissen

Andresdev02:
## Description

New toggle button to show only active mandatarissen. 

## How to test

Toggle the button and check if the results are only active mandatarissen.

## Links to other PR's

/

### DIFF
--- a/app/routes/organen/orgaan/mandatarissen.js
+++ b/app/routes/organen/orgaan/mandatarissen.js
@@ -5,12 +5,14 @@ import { service } from '@ember/service';
 import { getFormFrom } from 'frontend-lmb/utils/get-form';
 import { MANDATARIS_NEW_FORM_ID } from 'frontend-lmb/utils/well-known-ids';
 import { foldMandatarisses } from 'frontend-lmb/utils/fold-mandatarisses';
+import moment from 'moment';
 
 export default class OrganenMandatarissenRoute extends Route {
   @service store;
   @service installatievergadering;
 
   queryParams = {
+    activeOnly: { refreshModel: true },
     filter: { refreshModel: true },
     page: { refreshModel: true },
     size: { refreshModel: true },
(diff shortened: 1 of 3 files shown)


app/routes/organen/orgaan/mandatarissen.js
```diff
@@ -77,4 +79,14 @@ export default class OrganenMandatarissenRoute extends Route {
 
     return queryParams;
   }
+
+  getFilteredMandatarissen(mandatarissen, params) {
+    let filteredMandatarissen = mandatarissen;
+    if (params.activeOnly) {
+      filteredMandatarissen = mandatarissen.filter((mandataris) =>
+        moment().isBetween(mandataris.foldedStart, mandataris.foldedEnd)
```
bfidlers, line 87: This isBetween returns false if foldedEnd is not set. But we do want to see those. 
You can first check if foldedEnd is set en then do this check.

bfidlers (changes requested):
If mandataris has no end, it should show as active.

app/routes/organen/orgaan/mandatarissen.js
bfidlers: or maybe better, use the isActive method on the mandataris model

app/routes/organen/orgaan/mandatarissen.js
```diff
@@ -77,4 +80,21 @@ export default class OrganenMandatarissenRoute extends Route {
...
+          );
+          return isActive ? fold : null;
+        })
+      );
+
+      return filteredMandatarissen.filter((fold) => fold !== null);
+    }
+
+    return mandatarissen;
+  }
```
bfidlers, line 99:
```suggestion
 getFilteredMandatarissen(mandatarissen, params) {
    let filteredMandatarissen = mandatarissen;
    if (params.activeOnly) {
      filteredMandatarissen = mandatarissen.filter((mandataris) =>
        mandataris.mandataris.isActive
      );
    }
    return filteredMandatarissen;
  }
```

bfidlers (changes requested):
Uw vorige attempt was eigenlijk heel goed, gewoon de check die ge gebruikte kon iets beter, zie suggestie.
Ge moet dan wel terug die service wegdoen en die await.

bfidlers (approved):
Great!

PR #296, opened 2024-08-02
LMB-642 fetch mandatarissen in mandaten as links based on bestuursperiode

bfidlers:
## Description

Fetch mandatarissen in mandataris search page with bestuursperiode. This way the mandaten as links shown in the table don't show mandates of other bestuursperiodes.

## How to test

Go to the mandataris search page of a few bestuurseenheden and see that there are nu duplicate mandates shown in the mandate column.


JonasVanHoof (commented):
Do we just want to show the mandaten the person has in the bestuursperiode or do we also only want to show the active mandaten in the bestuurtsperiode?

JonasVanHoof (approved):

PR #297, opened 2024-08-02
Ben/lmb 634 fractie niet beschikbaar

bfidlers:
## Description

Show fractie niet beschikbaar instead of onafhankelijk in the mandataris tables. If multiple fracties currently linked for multiple mandates, all will be shown. If you are onafhankelijk and had another fractie before, this will be shown between brackets.

## How to test

Just go to a few bestuurseenheden, check if the fracties shown are correct. 
In a table within a bestuursorgaan it is easy, only one mandataris can exist there.
In the mandataris search page, it can be possible that there exist two mandatarissen with different fracties for the same person.


JonasVanHoof (approved):
Seems to work ok, what do we do with the fractie filter?

PR #306, opened 2024-08-22
make publication status editable again

Rahien:
## Description

Makes publication status editable again (the select got removed somehow)
image

image

image


bfidlers (approved):
Works like a charm!

PR #309, opened 2024-08-27
fix selecting persons: filter in backend, not afterwards in frontend

Rahien:
## Description

fix selecting persons: filter in backend, not afterwards in frontend. Doing this breaks pagination and may give the impression that there are no matches when there are only non-elected persons on the first page.

## How to test

Add a gemeenteraadslid to a current bestuursorgaan of type gemeenteraad. Search for the character "a". It should give you a full page of results.

## Links to other PR's
Needs this backend PR and a rebuild from a toLoad folder in the backend to work
- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/193

### DIFF
--- a/app/components/person/search-form.js
+++ b/app/components/person/search-form.js
@@ -80,6 +80,21 @@ export default class SharedPersoonPersoonSearchFormComponent extends Component {
       return;
     }
 
+    const extraFilter = {};
+    if (this.args.searchElected && this.args.bestuursperiode) {
+      extraFilter.verkiezingsresultaten = {
+        kandidatenlijst: {
+          verkiezing: {
+            'bestuursorgaan-in-tijd': {
+              'heeft-bestuursperiode': {
+                ':id:': this.args.bestuursperiode.id,
+              },
+            },
+          },
+        },
+      };
+    }
+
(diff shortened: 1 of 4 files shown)


app/components/person/search-form.js
```diff
@@ -110,12 +126,6 @@ export default class SharedPersoonPersoonSearchFormComponent extends Component {
     } catch (e) {
       this.error = true;
     }
-    if (this.args.searchElected && this.args.bestuursperiode) {
-      return this.verkiezingService.getPeopleThatAreElected(
```
bfidlers, line 114: This `getPeopleThatAreElected` is now only used with an array of just one person. So I would suggest renaming that function that something that makes more sense like `isElected` and simplify the logic a bit.

bfidlers (approved):
Makes more sense than before, looks good and works as expected. I would suggest rewriting some other stuff this influences (see comment).

app/components/person/search-form.js
Rahien: makes sense!

bfidlers (approved):
Perfect, the leftover function makes much more sense now.

PR #313, opened 2024-08-27
fix inverse relation verkiezing

bfidlers:
## Description

Fix ember resouce issue.

## How to test

Just look at the model and check if the inverse relations are okay.


JonasVanHoof (approved):
Nice one

PR #315, opened 2024-09-02
keep track on when to update bcsd alert

Rahien:
## Description

This feels a bit like premature optimization BUT i added a service that tracks whether we should refresh our calls to the backend to determine whether the voorzitter of the BCSD should be updated. 
The task with a timeout is in fact a pretty understanable and straight forward way of handling this that avoids passing all kinds of data around. 
The optimization I did was keeping a Date in a service so that you know if your data should still be up to date.
## How to test

Add a voorzitter to the BCSD, see the alert pop up.
Add a person to the gemeenteraad, no change to the alert.
Add the voorzitter to the ocwm, see the alert be removed.
remove the voorzitter from the ocwm, see the alert pop up again
Add the voorzitter as a member of the college of burgemeesters and schepenen, alert is still there.
Copy the college to the vast bureau, the alert should be removed.

Always git it 10 seconds for the alert to update its state.

### DIFF
--- a/app/components/mandaat/burgemeester-selector.js
+++ b/app/components/mandaat/burgemeester-selector.js
@@ -20,6 +20,7 @@ import { getFormFrom } from 'frontend-lmb/utils/get-form';
 
 export default class MandaatBurgemeesterSelectorComponent extends Component {
   @service store;
+  @service bcsd;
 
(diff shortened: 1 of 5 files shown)


bfidlers (approved):
Works as expected.

PR #316, opened 2024-09-06
beeindigd is not a real status. It exists only for the user interface and simply sets the end date for the user

Rahien:


JonasVanHoof (approved):
Okido 🚀

PR #317, opened 2024-09-06
fix form not scrollable

Rahien:
## Description

The form was not scrollable in case the screen was small
## How to test

Make your screen small but wide enough to still have a sidebar it should look like this
image

example url: http://localhost:4200/organen/15f48ffbd10856140a84e744b82c0835a29970902210f556fa5692b103c7590e/mandataris/new?person=ac7c0cec477cc285a64e2ab49dbf472d72552c43caf849a827ac7766b9894282

(go to a person and add a mandate there)


JonasVanHoof (approved):
Looks good! 
image

PR #319, opened 2024-09-09
LMB-724: when no publication status set to "Niet beschikbaar"

JonasVanHoof:
## Description

Old mandatarissen that have no publciatie status need to be set to "Niet beschikbaar" instead of "Bekrachtigd"

## How to test

1. Look at a mandataris in bestuursperiode 2012 - 2019 => iet beschikbaar
2. Create a new mandataris => status should be draft
3. Change status to bekrachtigd => pill turns green

### DIFF
--- a/app/components/mandaat/publicatie-status-pill.hbs
+++ b/app/components/mandaat/publicatie-status-pill.hbs
@@ -1,10 +1,7 @@
 <div>
   <AuPill
     @skin={{if
-      (or
-        (not @mandataris.publicationStatus.label)
-        (eq @mandataris.publicationStatus.label "Bekrachtigd")
-      )
+      (eq @mandataris.publicationStatus.label "Bekrachtigd")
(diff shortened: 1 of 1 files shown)


bfidlers (approved):
Looks good, it is a bit inconsistent however with the regular status. That one does not have a pill on "Niet beschikbaar". I think both should be handled the same way. 
Either both with a pill or both without.

app/components/mandatarissen/mandataris-card.hbs
```diff
@@ -16,7 +16,9 @@
             {{this.status}}
           </AuPill>
         {{else}}
-          <p>Niet beschikbaar</p>
+          <AuPill>
+            Niet beschikbaar
+          </AuPill>
```
JonasVanHoof, line 21: Updated it to a pill 
image

PR #320, opened 2024-09-09
LMB-732 | switch relation fractie kandidaatlijst

JonasVanHoof:
## Description

Switch relation on kandidaatlijst and fractie

## How to test

Split a fractie in legislatuur, fractie should have a kandidaatlijst

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/207

### DIFF
--- a/app/models/fractie.js
+++ b/app/models/fractie.js
@@ -26,7 +26,6 @@ export default class FractieModel extends Model {
 
   @belongsTo('kandidatenlijst', {
     async: true,
-    inverse: 'resulterendeFracties',
   })
   origineleKandidatenlijst;
 


bfidlers (approved):
Looks good, small comment about the ldes-streams in the backend.

bfidlers:
Bit confused, what does deleting this change? These two models are still linked with the same relation no? Does it matter if the inverse is mentioned here?

JonasVanHoof:
> Bit confused, what does deleting this change? These two models are still linked with the same relation no? Does it matter if the inverse is mentioned here?

The model is not reflecting the domain in the app anymore if we leave that, right?

PR #322, opened 2024-09-10
LMB-743 | Bug form change triggers unnecassary network calls

JonasVanHoof:
## Description

When testing on QA the correct form is not working smoothly. (Cannot edit a date without a refresh of the form)
Updated it that when the form observer is triggered the components used will only execute there logic when a specific predicate is changed (inserted or deleted).

## How to test

Create/wijzig/edit a mandataris and see if it is hard or tough to change a value.

### DIFF
--- a/app/components/rdf-input-fields/mandataris-rangorde.js
+++ b/app/components/rdf-input-fields/mandataris-rangorde.js
@@ -12,6 +12,8 @@ import {
   triplesForPath,
   updateSimpleFormValue,
 } from '@lblod/submission-form-helpers';
+import { isPredicateInObserverChange } from 'frontend-lmb/utils/is-predicate-in-observer-change';
+import { MANDATARIS_PREDICATE } from 'frontend-lmb/utils/constants';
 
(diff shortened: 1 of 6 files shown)


app/components/rdf-input-fields/mandataris-rangorde.js
```diff
@@ -42,7 +44,16 @@ export default class RdfMandatarisRangorde extends InputFieldComponent {
...
+      const mustTrigger = isPredicateInObserverChange(
+        formChange,
+        MANDATARIS_PREDICATE.mandaat
+      );
+
+      if (mustTrigger) {
+        onFormUpdate();
+      }
+    });
     onFormUpdate();
```
JonasVanHoof, line 57: Should we make it a method outside the `registerObserver()` method?

bfidlers: It could be abstracted a bit, but not really sure that is worth it.

bfidlers (approved):
Couldn't really seem to reproduce the issue, but it still works, so seems good.

PR #323, opened 2024-09-10
add possibility for admin to impersonate users

Rahien:
## Description

add possibility for admin to impersonate users

Note: the icons are slightly different in loket, this is because they use a more recent appuniversum version. Updating seems to require bumping ember major version, which merits its own story.
## How to test

Log in as Agentschap Binnenlands Bestuur, click the impersonate icon in the top bar:

image

image

Select another user:
image

See that you can see what this user sees

image

End the impersonation session:
image

## Links to other PR's
Requires these changes on the app side to work:
- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/211


Rahien:
> Works as a charm!
> 
> Should we hide the modules on the main page when the admin is logged in? They are clickable now but not doing anything atm

we can beautify later :)

PR #325, opened 2024-09-10
LMB-630 | Set replacement when verhinderd

JonasVanHoof:
## Description

You need to be able to set a replacement when the mandataris is verhinderd, even when you did not slect a replacement when wijziging the mandataris.

## How to test

Set status to verhinderd, do not select a replacement. Set a replacement in the mandataris card.


app/models/mandataris-status-code.js
```diff
@@ -3,4 +3,8 @@ import Model, { attr } from '@ember-data/model';
 export default class MandatarisStatusCodeModel extends Model {
   @attr label;
   @attr uri;
+
+  get isVerhinderd() {
```
bfidlers, line 7: This might give some issues with titelvoerend, which is the burgemeester variant of verhinderd.

app/components/mandatarissen/mandataris-card.js
```diff
@@ -51,4 +59,24 @@ export default class MandatarisCardComponent extends Component {
...
+        publicationStatus: await getDraftPublicationStatus(this.store),
+      }
+    );
+    const lidmaatschap = await this.args.mandataris.heeftLidmaatschap;
+    const replacementMandataris =
+      await this.mandatarisService.getOrCreateReplacement(
+        this.args.mandataris,
+        replacement,
+        newMandatarisProps,
+        await lidmaatschap.binnenFractie
```
bfidlers, line 77: I think we really need to rethink these fracties of replacements, but could be a refinement story or something like that?
I think it goes wrong here somewhere if the person you are trying to replace has no fractie:
Screenshot 2024-09-20 at 10 29 07

app/components/mandatarissen/mandataris-card.hbs
```diff
@@ -103,24 +103,39 @@
...
-              }}
-            </AuLink>
-
-          {{/each}}
+        <div class="au-c-description-value replacement">
+          {{#if (not (await this.vervangersDoor))}}
+            <Mandatarissen::ReplacementSelect
+              @showDelete={{false}}
+              @showEditAsIcon={{true}}
+              @nothingSelectedText="Geen vervanger geselecteerd, voeg er een toe door op het potloodje te klikken."
```
bfidlers, line 119: Do we really need the "voeg er een toe door op het potloodje te klikken"? Feels a bit over the top.

bfidlers (changes requested):
Few things go wrong for me. 
If you look at a burgemeester mandate, replacements are no longer shown, because a burgemeester is not "verhinderd" but "titelvoerend". 
Also it errors sometimes, I think when the mandatee in question has no fractie, I mentioned it in a comment in the code.
UI goes a bit haywire when you try to replace a replacement. See image below:
Screenshot 2024-09-20 at 10 46 02

app/models/mandataris-status-code.js
bfidlers: Shouldn't we also check the concept uri instead of trying to match the label?

app/components/mandatarissen/mandataris-card.hbs
JonasVanHoof: Yes Saska already mentioned this in the Jira ticket but dind't update it it yet :D

app/models/mandataris-status-code.js
JonasVanHoof: Updated it to check on URI of verhinderd and titelvoerend

app/components/mandatarissen/mandataris-card.hbs
JonasVanHoof: "Voeg vervanger toe"

app/components/mandatarissen/mandataris-card.hbs
```diff
@@ -103,24 +103,43 @@
...
+    {{#if (await @mandataris.status.isVerhinderd)}}
+      {{#if
+        (not (or (await this.vervangersVan) @mandataris.bekleedt.isSchepen))
+      }}
+        <div class="au-o-grid__item au-u-1-2">
+          <div class="au-c-description-label"></div>
+          <div class="au-c-description-value au-u-margin-top-tiny">
+          </div>
+        </div>
+      {{/if}}
```
JonasVanHoof, line 115: Fixes the haywire element

 GR: 
image

GR met dubbele vervanger:
image


Schepenen:
image

Schepenen met dubbele vervanger:
image

app/components/mandatarissen/mandataris-card.js
JonasVanHoof: Very true did have that issue to! We should flag or limit functionality when not all information is set.

PR #329, opened 2024-09-19
fix broken page installatievergadering ocmw

bfidlers:
## Description

Fix broken page installatievergadering OCMW. This page did error because it tried to read some model info, the page did not have in the case the bestuurseenheid was an OCMW.

## How to test

Go to an installatievergadering page of an OCMW and check it does not error anymore.


JonasVanHoof (approved):
Seems to work  fine, from empty page and console error to  this :)
image

PR #330, opened 2024-09-24
lmb-700 | limit the start and endate input iwhen creating a new mandataris 

JonasVanHoof:
## Description

When creating a new mandataris the start and end date need to be validated that they are in the bestuursperiode of the bestuursorgaan in tijd. Added a validation for the date field that with the help from our application context can valdiate of the value in the input field is correct.

## How to test

1. Create a mandataris and insert a date that is not in the current bestuursperiode. 
2. Add a date that is exact the start or end date of the bestuursorgaan bindingStart or einde date

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/223

### DIFF
--- a/app/components/mandatarissen/update-state.hbs
+++ b/app/components/mandatarissen/update-state.hbs
@@ -25,19 +25,19 @@
     <AuLabel>
       Vanaf
     </AuLabel>
-    <AuDatePicker
+    <AuDateInput
       @value={{this.date}}
       @onChange={{this.updateDate}}
-      @error={{not this.validDate}}
-      @alignment="top"
+      @error={{not this.isValidDate}}
+      @alignment="default"
     />
-    {{#unless this.validDate}}
+    {{#unless this.isValidDate}}
       <AuAlert
         @size="small"
         @closable={{false}}
         @skin="error"
         class="au-u-margin-top"
-      >De gekozen datum moet na de startdatum liggen van het huidige mandaat.</AuAlert>
+      >{{this.invalidDateErrorMessage}}</AuAlert>
     {{/unless}}
   </div>
   <div>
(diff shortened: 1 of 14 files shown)


bfidlers (changes requested):
Looks mostly good, but I'm having a few issues:
1. The corrigeer fouten seems to error whenever I try to open it for a burgemeester
Screenshot 2024-09-24 at 17 02 36
2. The update state also needs an update, now (was already like this) it only checks if the start date is after the start date that was entered before, but you also need to check if it is not past the end of the bestuursperiode
Screenshot 2024-09-24 at 17 05 51

JonasVanHoof:
@bfidlers 

1. More than one bestuursorgaan was passed here and than it returned the value null, now it will always get the first one 
2. Indeed I forgot about this one.. 
 - changed the input from date picker to a dateInput
 - added a check that the madnataris start date cannot be out of the bestuursperiode
 -  update the error message so it is clear why the error is shown (in the past or after bestuursperiode einde)
 -  added a reference date as the user was not able to fill in the current date of the mandataris again ( when opening the modal with date 24-09-2024 and than changing it to 25-09-2024 and than back to 24-09-2024 this gave an error because of the hours where not correct)
 3. Other:
 - fixed the loading state of the pas aan button
 - check if date is also changed in HasChanges() method
 - renamed the method to _getApplicationContextMetaTtl_

bfidlers (approved):
Noticed a few issues with the corrigeer fouten, where it sometimes does not work, however I found the same issue on master (I think it has something to do with the fractie handling), logged a ticket for that https://binnenland.atlassian.net/jira/software/c/projects/LMB/boards/225/backlog?selectedIssue=LMB-800, but I think this PR does not cause this fault and otherwise it looks good.

PR #333, opened 2024-10-01
automatically create bestuursorgaan and mandates from template

Rahien:
## Description

- auto create mandates for bestuursorgaan
- use the proper, new conceptscheme for non standard bestuursorgaan instances
- have the forms wait on their onSave and onCreate promise so they don't seem unresponsive

## How to test

Go to the interface to create new bestuursorgaan, use it, see that the mandates have automatically been added.

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/228


bfidlers (approved):
Looks good.

Noticed a few things during testing, which are not really an issue with this story, but should be looked into.
In the selector we only get to see 20 options I think, but this can be confusing if you don't know which options there are and it is not clear that there are more options than shown.
Also it is not possible to add new people to the bestuursorganen that can be added. They are restricted to the kieslijst.

Rahien:
> Looks good.
> 
> Noticed a few things during testing, which are not really an issue with this story, but should be looked into. In the selector we only get to see 20 options I think, but this can be confusing if you don't know which options there are and it is not clear that there are more options than shown. Also it is not possible to add new people to the bestuursorganen that can be added. They are restricted to the kieslijst.

thanks! we created separate issues for those!

PR #336, opened 2024-10-03
allow ending mandate if state is not set, fix beëindigd trema

Rahien:
## Description

allow ending mandate if state is not set, fix beëindigd trema

## How to test

End a mandataris that has no state set, e.g. in boutersem :)


bfidlers (approved):
Looks good!

PR #338, opened 2024-10-04
better mandataris statuses

Rahien:
## Description
Better statuses 
image

## How to test

Boutersem has many statusses and is a nice test


app/components/mandaat/publicatie-status-pill.hbs
```diff
@@ -3,7 +3,11 @@
     @skin={{if
       (eq @mandataris.publicationStatus.label "Bekrachtigd")
       "success"
-      "default"
+      (if
```
Rahien, line 6: i was going to rework this like in the mandataris-status-pill BUT ran into the fact that the publication status is not always loaded and didn't want to deal with promise hell

bfidlers (commented):
In general it is a nice improvement, but I'm having some issues with it:

- the sorting seems kind of scuffed, not sure waht it is sorting on, but it does not work entirely. 
- Not a big fan of the beeindigd status, does it really matter what the state before was? Can't we just show beeindigd in red or something like that and if they want to now more, they can click on the mandate and get the info there

app/components/organen/mandataris-table.hbs
```diff
@@ -78,12 +83,19 @@
           {{row.mandataris.bekleedt.bestuursfunctie.label}}
         </LinkTo></td>
       <td>
-        {{row.mandataris.status.label}}
+        <Mandaat::MandatarisStatusPill @mandataris={{row.mandataris}} />
```
bfidlers, line 86: Shouldn't this have the `@detailView=false`? Otherwise that argument is never used/

app/templates/mandatarissen/persoon/mandaten.hbs
```diff
@@ -127,12 +130,15 @@
...
-          <td>
+          <td class={{if (is-in-past row.foldedEnd) "au-u-muted"}}>
             {{moment-format row.foldedStart "DD-MM-YYYY"}}
           </td>
-          <td>
+          <td class={{if (is-in-past row.foldedEnd) "au-u-muted"}}>
             {{moment-format row.foldedEnd "DD-MM-YYYY"}}
           </td>
+          <td>
+            <Mandaat::MandatarisStatusPill @mandataris={{row.mandataris}} />
```
bfidlers, line 140: Same as comment above

bfidlers (commented):
Seems like the detailView argument should either be used in the mentioned places or just removed, because no there is no differentiation.
Also still some issues with sorting, not sure if this is by design? Because it is an improvement and it doesn't really bother me like this.
Screenshot 2024-10-08 at 12 13 08

app/components/organen/mandataris-table.hbs
bfidlers: huh weird in your example it does show the beeindigd in the overview, for me it shows the niet beschikbaar, seems like it defaults to true for you and false for me if not mentioned, weird.

Rahien: this is a screenshot from an earlier version, your comments also seem to refer to an earlier version... i've pushed everything, did your pull come through correctly?

Rahien: detailView is falsy by default, and only true for the mandataris card

Rahien:
> Seems like the detailView argument should either be used in the mentioned places or just removed, because no there is no differentiation. Also still some issues with sorting, not sure if this is by design? Because it is an improvement and it doesn't really bother me like this.
Fixed sorting! Good shout!

app/components/organen/mandataris-table.hbs
bfidlers: didn't have the latest changes indeed

bfidlers (approved):
Looks good!

PR #340, opened 2024-10-04
Fix comparing dates in update state form

bfidlers:
## Description

Update state form could always be saved because of an issue with comparing wrong dates.

## How to test

Go to the update state form and check if the form is disabled and gets enabled when you update the date.

### DIFF
--- a/app/components/mandatarissen/update-state.js
+++ b/app/components/mandatarissen/update-state.js
@@ -152,7 +152,7 @@ export default class MandatarissenUpdateState extends Component {
   get isInputDateTheSameAsMandatarisStart() {
     const startDate = new Date();
     const inputDate = new Date(this.date);
-    return startDate.getTime() === inputDate.getTime();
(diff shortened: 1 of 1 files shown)


app/components/mandatarissen/update-state.js
```diff
@@ -150,11 +150,8 @@ export default class MandatarissenUpdateState extends Component {
   }
 
   get isInputDateTheSameAsMandatarisStart() {
-    const startDate = new Date(this.args.mandataris.start);
-    startDate.setHours(0, 0, 0, 0);
+    const startDate = new Date();
     const inputDate = new Date(this.date);
-    inputDate.setHours(0, 0, 0, 0);
-
     return startDate.getTime() === inputDate.getTime();
```
Rahien, line 155: isn't it possible that this fails if we're right on the boundary of the millisecond? I agree that the chance is small... Maybe we should do something like the code before tries to do, maybe using momentjs?

Rahien (changes requested):

PR #341, opened 2024-10-07
fixes for iv page

Rahien:
## Description

Make the installatievergadering page update when the burgemeester is changed.
Show warnings about burgemeester in the college orgaan the same way as it is done for bcsd

## How to test

Change the mayor, see the college and vast bureau roles change
Add a true mayor, see an error
Add an extra aangewezen burgemeester, see an error

### DIFF
--- a/app/components/mandaat/burgemeester-selector.js
+++ b/app/components/mandaat/burgemeester-selector.js
@@ -169,6 +169,10 @@ export default class MandaatBurgemeesterSelectorComponent extends Component {
     );
     this.setup.perform();
     this.bcsd.forceRecomputeBCSD();
+    if (this.args.onUpdateBurgemeester) {
+      this.args.onUpdateBurgemeester();
+    }
+    
   });
 
   @action
--- a/app/components/mandaat/burgemeester-selector.js
+++ b/app/components/mandaat/burgemeester-selector.js
@@ -1,10 +1,11 @@
 import Component from '@glimmer/component';
 
+import { action } from '@ember/object';
 import { service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
-import { action } from '@ember/object';
 
-import moment from 'moment';
+import { restartableTask } from 'ember-concurrency';
+import { getFormFrom } from 'frontend-lmb/utils/get-form';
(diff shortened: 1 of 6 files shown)


app/components/mandaat/burgemeester-selector.js
```diff
@@ -17,14 +17,14 @@ import {
 import { toUserReadableListing } from 'frontend-lmb/utils/to-user-readable-listing';
```
JonasVanHoof, line 17: Needs to be removed for lint job to succeed*

app/components/mandaat/burgemeester-selector.js
```diff
@@ -17,14 +17,14 @@ import {
...
 import { restartableTask } from 'ember-concurrency';
 import { getFormFrom } from 'frontend-lmb/utils/get-form';
+import { CBS_BESTUURSORGAAN_URI } from 'frontend-lmb/utils/well-known-uris';
```
JonasVanHoof, line 20: Needs to be removed for lint job to succeed*

app/components/verkiezingen/burgemeester-alert.js
```diff
@@ -0,0 +1,90 @@
...
+  @tracked lastRecomputeTime = null;
+
+  constructor() {
+    super(...arguments);
+    this.bcsd.forceRecomputeBCSD();
+  }
+
+  get collegeOrgaanInTijd() {
+    return this.args.collegeBestuursorgaanInTijd;
+  }
```
JonasVanHoof, line 25: Not really necessary imo

JonasVanHoof (approved):
Looks good, do we also need to add a warning that they have added two aangewezen burgemeesters?
image

app/components/verkiezingen/burgemeester-alert.js
Rahien: i sortof like this pattern for documentation purposes, this way you see better that there is an arg

Rahien:
> Looks good, do we also need to add a warning that they have added two aangewezen burgemeesters? 

it's there but the burgemeester one takes prio

PR #342, opened 2024-10-07
fix bestuurseenheid contact created in wrong direction, triggering mu-auth issue

Rahien:
## Description

we're not allowed to modify bestuurseenheden! So we need to create the contact instance and link from it to bestuurseenheid to be allowed by mu-auth

## How to test

open the settings page and set the email for a bestuurseenheid

## Links to other PR's

- requires app change: https://github.com/lblod/app-lokaal-mandatenbeheer/pull/236


JonasVanHoof (approved):
Great job

PR #346, opened 2024-10-07
Lmb 826 make add message optional in change modal

JonasVanHoof:
## Description

The modal that adds a description to your change does not need to be shown on the legislatuur page.

## How to test

Correct a mandataris. The modal should show
Edit a mandataris in the IV page the modal should not show.

### DIFF
--- a/app/components/form/save-with-history.hbs
+++ b/app/components/form/save-with-history.hbs
@@ -1,36 +1,30 @@
-{{#if @isModalOpen}}
-  {{#if @addDescription}}
-    <AuModal
-      @id="form-history"
-      @closeModal={{fn (mut @isModalOpen) false}}
-      @modalOpen={{@isModalOpen}}
-      class="small"
-      @title="Beschrijving doorgevoerde wijzigingen."
-      as |Modal|
-    >
-      <Modal.Body>
-        <AuInput
-          @id="history-message"
-          @width="block"
-          placeholder="Beschrijving wijzigingen"
-          {{on "input" @updateValue}}
-        />
-      </Modal.Body>
-      <Modal.Footer>
-        <AuToolbar as |Group|>
-          <Group>
-            <AuButton
-              {{on "click" @save}}
-              @loading={{@isSaving}}
-              @loadingMessage="Aan het opslaan"
-            >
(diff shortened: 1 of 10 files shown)


app/components/mandatarissen/mandataris-extra-info-card.hbs
```diff
@@ -43,6 +43,7 @@
 >
   <div class="au-o-box">
     <Form::Instance
+      @showHistoryDescription={{true}}
```
Rahien, line 46: i would make true the default

app/components/form/save-with-history.hbs
```diff
@@ -1,30 +1,36 @@
...
-          @loadingMessage="Aan het opslaan"
-        >
-          Bevestig
-        </AuButton>
-      </Group>
-    </AuToolbar>
-  </Modal.Footer>
-</AuModal>
\ No newline at end of file
+{{#if @isModalOpen}}
```
Rahien, line 1: this if seems redundant?

app/components/form/save-with-history.hbs
```diff
@@ -1,30 +1,36 @@
...
+              @loadingMessage="Aan het opslaan"
+            >
+              Bevestig
+            </AuButton>
+          </Group>
+        </AuToolbar>
+      </Modal.Footer>
+    </AuModal>
+  {{else}}
+    <code {{did-insert @save}} />
```
Rahien, line 34: this doesn't feel quite right. Shouldn't the instance component just do this? and either trigger the saveWithHistory or not depending on @showHistoryDescription?

Rahien (changes requested):

app/components/mandatarissen/mandataris-extra-info-card.hbs
JonasVanHoof: I actually thought you would say it ;p I'll change it!

app/components/form/save-with-history.hbs
JonasVanHoof: Ill update it that it is being handled  in a higher layer

Rahien (approved):

PR #348, opened 2024-10-07
fix restore fractie failing silently

Rahien:
## Description

When the destroy of a fractie fails, it failed silently. now we show an error toast and don't remove the fractie from the list.
## How to test

Weirdly i could trigger this by writing and extra s in the type of my system-notification has-many relation, making it system-notification-links instead of system-notification-link as a type.


bfidlers (changes requested):
Could not trigger the failing of reverting the fracties.
Also this seems to break the original functionality. 
Now for a split kieslijst, only the one you selected is put back.

Rahien:
> Could not trigger the failing of reverting the fracties. Also this seems to break the original functionality. Now for a split kieslijst, only the one you selected is put back.

Fixed only the first one being put back.
You could also trigger a fail by setting throttling to 'offline' in the network tab right before you do this

bfidlers (approved):
Looks good!

PR #349, opened 2024-10-07
lmb-830: show all bestuursorgaan classificatie codes when adding a bestuursorgaan

JonasVanHoof:
## Description

We only show 20 items in the conceptscheme selector. Raise the limit of the page to 9999 so it shows all the items. 

## How to test

Add a new bestuursorgaan and check that it shows more than 20 results

## Attachments

image


JonasVanHoof:
Notice that when searching for items the options are not reloaded again when clearing the search ⚠️

PR #351, opened 2024-10-08
Lmb 856 | Rework date picker use it everywhere

JonasVanHoof:
## Description

TODO

## How to test

TODO

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/180


app/components/date-input.js
```diff
@@ -0,0 +1,94 @@
...
+
+    this.args.onChange?.(date);
+  });
+
+  isDateInRange(date, min, max) {
+    if (!date) {
+      return false;
+    }
+
+    date.setHours(0, 0, 0, 0);
```
Rahien, line 44: would rework this so it uses moment, this setHours thing got us in trouble before in another story

Rahien (changes requested):
just the setHours thing, use moment it will do all this better than we can do it!

app/components/form/instance.js
```diff
@@ -44,6 +44,14 @@ export default class InstanceComponent extends Component {
...
     this.errorMessage = null;
 
+    const isValid = isValidForm(this.formInfo);
+    this.forceShowErrors = !isValid;
+    if (!isValid) {
+      this.errorMessage =
+        'Niet alle velden zijn correct ingevuld. Gelieve deze eerst te corrigeren.';
+      this.save.cancelAll();
+    }
+
```
JonasVanHoof, line 54: This was a freaking pain to find..

app/components/date-input.js
JonasVanHoof: Done!

Rahien (approved):

app/components/mandatarissen/update-state.hbs
```diff
@@ -22,23 +22,15 @@
...
-    </AuLabel>
-    <AuDateInput
+    <DateInput
+      @label="Vanaf"
       @value={{this.date}}
-      @onChange={{this.updateDate}}
-      @error={{not this.isValidDate}}
-      @alignment="default"
+      @from={{this.today}}
+      @to={{@mandataris.einde}}
```
Rahien, line 29: i think indeed this is not ok and it should be limited to the start and end of the period

Rahien (changes requested):

app/components/mandatarissen/update-state.hbs
Rahien: (maybe the mandataris doesn't have an end date)

app/components/rdf-input-fields/rdf-date-input.hbs
```diff
@@ -0,0 +1,27 @@
...
+  for={{this.inputId}}
+  @error={{this.hasErrors}}
+  @warning={{this.hasWarnings}}
+  @required={{this.isRequired}}
+>{{this.title}}</AuLabel>
+<DateInput
+  @id={{this.inputId}}
+  @value={{this.date}}
+  @onChange={{this.onUpdate}}
+  @showErrorAsAlert={{false}}
```
Rahien, line 11: i would expect this one to also be restricted to the current period

app/components/mandatarissen/update-state.hbs
JonasVanHoof: Isn't it weird than that we are filling in todays date  as default  when opening the wijzig mandataris?

JonasVanHoof: Yes de date input handles a null date as enddate

app/components/rdf-input-fields/rdf-date-input.hbs
JonasVanHoof: These restrictions are still set in the ttl as a validations *

Rahien (approved):

PR #366, opened 2024-10-11
Lmb 812 hide iv for districts

JonasVanHoof:
## Description

Districts should never see the legislatuur module., they prepare it in the normal interface.

## How to test

1. does not see the menu item (main and compact menu)
2. on mandataris search page the period is not disabled
3. on organen route the `beheer fracties` are enabled
4. the mandatarissen of 2025 - heden can be edited

When in another bestuurseenheid than district the legislatuur should work as before

### DIFF
--- a/app/controllers/mandatarissen/persoon/mandataris.js
+++ b/app/controllers/mandatarissen/persoon/mandataris.js
@@ -113,15 +113,12 @@ export default class MandatarissenPersoonMandatarisController extends Controller
   });
 
   get isDisabledBecauseLegislatuur() {
-    if (
+    return (
       this.periodeHasLegislatuur &&
       this.behandeldeVergaderingen &&
-      this.behandeldeVergaderingen.length === 0
-    ) {
-      return this.model.isDistrictEenheid ? false : true;
-    }
-
-    return false;
+      this.behandeldeVergaderingen.length === 0 &&
+      !this.model.isDistrictEenheid
+    );
   }
 
   get toolTipText() {
--- a/app/controllers/mandatarissen/persoon/mandataris.js
+++ b/app/controllers/mandatarissen/persoon/mandataris.js
@@ -118,7 +118,7 @@ export default class MandatarissenPersoonMandatarisController extends Controller
       this.behandeldeVergaderingen &&
       this.behandeldeVergaderingen.length === 0
     ) {
-      return true;
+      return this.model.isDistrictEenheid ? false : true;
     }
 
     return false;
--- a/app/routes/mandatarissen/persoon/mandataris.js
+++ b/app/routes/mandatarissen/persoon/mandataris.js
@@ -12,7 +12,6 @@ import RSVP from 'rsvp';
 export default class MandatarissenPersoonMandatarisRoute extends Route {
   @service currentSession;
   @service store;
-  @service currentSession;
 
   async model(params) {
     const bestuurseenheid = this.currentSession.group;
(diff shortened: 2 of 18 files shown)


app/controllers/mandatarissen/persoon/mandataris.js
```diff
@@ -104,7 +104,7 @@ export default class MandatarissenPersoonMandatarisController extends Controller
       this.behandeldeVergaderingen &&
       this.behandeldeVergaderingen.length === 0
     ) {
-      return true;
+      return this.model.isDistrictEenheid ? false : true;
```
Rahien, line 107: why not add this into the if then?

app/routes/mandatarissen/persoon/mandataris.js
```diff
@@ -37,6 +39,7 @@ export default class MandatarissenPersoonMandatarisRoute extends Route {
       mandatarisExtraInfoForm,
       bestuursorganen,
       selectedBestuursperiode,
+      isDistrictEenheid: isDistrict,
```
Rahien, line 42: we can just make this a property of the session then maybe?

Rahien: same for other cases of this pattern

app/services/current-session.js
```diff
@@ -85,6 +87,20 @@ export default class CurrentSessionService extends Service {
     return this.roles.includes(role);
   }
 
+  async isDistrict() {
+    const classificatie = await this.group?.classificatie;
```
Rahien, line 91: if we fetch this during load, this can be just a getter

Rahien (changes requested):

Rahien:
I'm adopting this pr while Jonas is away :)

app/services/current-session.js
```diff
@@ -44,6 +45,7 @@ export default class CurrentSessionService extends Service {
         reload: true,
       });
       this.groupClassification = await this.group.classificatie;
+      this.isDistrict = await this.isDistrictBestuurseenheid();
```
JonasVanHoof, line 48: This is indeed a better approach 💪🏻

JonasVanHoof (commented):
Looks good visibility wise But i cannot seem to  open up the beheer fracties button? It works on master, DEV and test but not in this PR :/

To go to the fracties we add the hash bestuursperiode id and this one looks ok so not sure what was changed here, any idea?

JonasVanHoof:
> Looks good visibility wise But i cannot seem to open up the beheer fracties button? It works on master, DEV and test but not in this PR :/
> 
> To go to the fracties we add the hash bestuursperiode id and this one looks ok so not sure what was changed here, any idea?

This is because we do not have any fracties set for the period, maybe?

bfidlers:
> > Looks good visibility wise But i cannot seem to open up the beheer fracties button? It works on master, DEV and test but not in this PR :/
> > To go to the fracties we add the hash bestuursperiode id and this one looks ok so not sure what was changed here, any idea?
> 
> This is because we do not have any fracties set for the period, maybe?

No that should not matter, I think it is because in the fractie page we don't allow selecting the bestuursperiod of a bestuursperiod with a installatievergadering that is not behandeld. I think we should remove the installatievergadering for bestuurseenheden that do not need one, with a migration.

Rahien (approved):
approved after review by jonas (who couldn't approve himself as he started the pr originally)

PR #367, opened 2024-10-13
remove frontend code from mirroring and move it to the mandataris service

Rahien:
## Description
remove frontend code from mirroring and move it to the mandataris service

## How to test

- Add some mandataris instances to gemeente, CBS and aangewezen burgemeester
- Add some mandatarises (to be removed automatically) to ocmw, VB
- click 'neem over van' button in each OCMW organ
- see that the old mandataris instances have been removed (with tombstones). The new mandataris instances for the OCMW organs are created with fracties and beleidsdomeinen. Once the mandataris instances are fixed and the IV has been completed, the link is also made between these mandataris instances
- now remove the mayor and select a new one. This should still update all of his corresponding roles in the other organs

image

## Links to other PR's
requires the work done in the mandataris service
- https://github.com/lblod/mandataris-service/pull/50

### DIFF
--- a/app/components/mandaat/burgemeester-selector.js
+++ b/app/components/mandaat/burgemeester-selector.js
@@ -109,9 +109,9 @@ export default class MandaatBurgemeesterSelectorComponent extends Component {
   }
 
   async loadBurgemeesterPersoon() {
-    const burgemeesters = await this.burgemeesterMandate.bekleedDoor;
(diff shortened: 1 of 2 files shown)


bfidlers (approved):
Works as expected.

bfidlers:
A thing to be aware of is that the new syncing also does not keep the order of the table it syncs, so we might need to think about another solution for sorting.

PR #368, opened 2024-10-14
Ben/lmb 821 no replacing yourself

bfidlers:
## Description

Don't allow users to add the same person as a replacement.
They check does happen in two places, in the replacement component itself and the components/templates that uses it. This is because the components are quite nested (person in replacement-select etc.) and it makes it easier to do validations etc this way.

There are a lot of changes in the `update-state.hbs` file, this is because I moved the modal there, since it easier to manipulate the modal title that way, and I added the person to the modal title to make it more clear which person you are editing.

## How to test

Try adding the same person as a replacement in the update state form.
Also try adding it when a mandataris is set to verhinderd but no replacement was selected.


app/components/mandatarissen/replacement-select.js
```diff
@@ -10,9 +10,14 @@ export default class MandatarissenReplacementComponent extends Component {
   @service mandataris;
 
   @tracked overlappingMandate;
+  @tracked sameMandate;
 
   checkMandate = task({ keepLatest: true }, async () => {
     const replacement = this.args.selected;
+    if (replacement.id === this.args.mandataris.isBestuurlijkeAliasVan.id) {
+      this.sameMandate = true;
```
Rahien, line 18: this doesn't seem to be used

Rahien (changes requested):

app/components/mandatarissen/replacement-select.js
bfidlers: Forgot to remove this indeed.

Rahien (approved):

PR #370, opened 2024-10-15
as bestuursperiode is now starting in december, use off by one to pick current bestuursperiode

Rahien:
## Description

I'm aware that this will not work between 04-12 and 31-12, but let's tackle that when we have to (if we have to)


bfidlers (approved):
If needed, we can always extend the bestuursperiode with an actual start and end date, or mark bestuursperiodes as current like you said.

I'm fine with leaving this for now and tackling this once the new period starts.

PR #371, opened 2024-10-16
don't allow replacing yourself in correct mistakes either

Rahien:
## Description
Filter out mandataris instances of same person when selecting replacement in correct mistakes

## How to test

obvious


bfidlers (approved):
Looks good.

PR #372, opened 2024-10-16
Ben/lmb 787 no independent fractie on create

bfidlers:
## Description

Restrict fractions when creating a mandataris. There are three cases where the fractie-selector is used:
- When creating a mandataris you are restricted to the current-fractie of the person for that bestuursperiod if it exists, if not all fracties (except independent) are possible.
- Correct mistakes allows all fracties, the independent one is the one set on the person if it exists, or a new one if it doesn't.
- Update state, allows the fracties of that mandataris in it's past states or independent (the same one for the person if it is already set in another mandataris of that person or a new one if it doesn't)

Also rewrote the entire fractie logic in the fractie-selector to simplify a bit.

## How to test

- Create a new mandataris in the prepare legislature, check if the fracties presented are correct.
- Create a new mandataris for the same person in a different organ and check if only the fractie earlier selected is available.
- Create a new mandataris in the current bestuursperiode that has no other mandate yet (all fracties except onafhankelijk should be possible).
- Create a new mandataris for the same person in the same bestuursperiode (only the previously selected fractie should be available)
- Also test the correct mistakes and update state logic of the the mandatarissen you just created.

## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/53


JonasVanHoof (approved):
Seems to be working 🚀

PR #375, opened 2024-10-17
Lmb 799 fix inconsistencies update state en corrigeer fouten forms

JonasVanHoof:
## Description

We want the wijzig mandataris and the corrigeer mandataris form to react and look as good as identical.

- Date input errors are shown as alerts in update-state
- No help text under the date input of update-state
- Status must have the **verplicht** plaque next to the label
- Fractie selector does not show **Verplicht** (field cannot be empty but also change it i assume for the looks)

## How to test

1. Fill in an invalid date in the wijzig mandataris modal date input and it should show the error as text
2. The helptext of the date should always be under the date input with the same value as the correct forms
3. Status selector should have required plaque
4. Fractie selector should have required plaque

## Attachments

image

// Helptext always shown
image

### DIFF
--- a/app/components/date-input.hbs
+++ b/app/components/date-input.hbs
@@ -14,6 +14,11 @@
     {{au-inputmask options=(hash mask="99-99-9999" placeholder="_")}}
     {{on "input" (perform this.onChange)}}
   />
+  {{#unless @showFormatHelpText}}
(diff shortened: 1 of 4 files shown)


JonasVanHoof:
Comparison

image

Rahien (approved):

PR #379, opened 2024-10-18
Lmb 599  | Automatic fillin of search form person info

JonasVanHoof:
## Description

Make sure that when searching for a person and it did not exist yet the data from the input fields  (first- lastname, rrn) is passed on to the create person form

## How to test

1. create eigen orgaan
2. add mandataris with search of existing person
3. add mandataris with search of NOT existing person
4. Create person see if the fields are prefilled

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/253

### DIFF
--- a/app/components/rdf-input-fields/rijksregister-input.js
+++ b/app/components/rdf-input-fields/rijksregister-input.js
@@ -25,7 +25,7 @@ export default class RDFRijksRegisterInput extends InputFieldComponent {
 
     if (matches.values.length > 0) {
       this.rijksregisternummer = matches.values[0].value;
-      this.removeEmptyBlankNodeFromAutoFillin();
+      this.removeEmptyBlankNodeFromPrefilledData();
       replaceSingleFormValue(this.storeOptions, this.rijksregisternummer);
     }
   }
@@ -38,7 +38,6 @@ export default class RDFRijksRegisterInput extends InputFieldComponent {
     const rijksregisternummer = event.target.value.trim();
     this.rijksregisternummer = rijksregisternummer;
 
-    this.removeEmptyBlankNodeFromAutoFillin();
     replaceSingleFormValue(
       this.storeOptions,
       rijksregisternummer ? rijksregisternummer : null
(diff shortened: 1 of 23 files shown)


app/components/rdf-input-fields/rijksregister-input.js
```diff
@@ -24,6 +25,8 @@ export default class RDFRijksRegisterInput extends InputFieldComponent {
 
     if (matches.values.length > 0) {
       this.rijksregisternummer = matches.values[0].value;
+      this.removeEmptyBlankNodeFromPrefilledData();
```
Rahien, line 28: i think that if you create a true uri + type + uuid you won't need this :)

app/utils/constants.js
```diff
@@ -38,8 +38,13 @@ export const placeholderNietBeschikbaar = {
 };
 
 export const MANDATARIS_PREDICATE = {
-  persoon: 'http://data.vlaanderen.be/ns/mandaat#isBestuurlijkeAliasVan',
-  mandaat: 'http://www.w3.org/ns/org#holds',
+  persoon: new Namespace('http://data.vlaanderen.be/ns/mandaat#')(
```
Rahien, line 41: why change this? i don't see the value in first wrapping it to then unwrap it again in the isPredicateInObserverChange.
I'm ok with it though, but i wouldn't have done this. If we can use simple strings we should use simple strings imo

Rahien (changes requested):

JonasVanHoof:
More issues from RC ⚠️

1) i entered a rrn and then switched to one without a date (the 00000 one), i couldn't edit the date anymore
2) i entered the fake rrn 06.02.02-000.80 and it assumes my person was born in 1906, i would assume 2006 and i would allow users to modify if we guessed wrong (unlikely)

ofc for endings like 87, still guess dates in the past, not in the future 🙂

tests/unit/utils/rijksregisternummer-test.js
```diff
@@ -0,0 +1,52 @@
...
+module(
+  'Unit | Utils | form-validations | RRN helpers | 06.02.02-000.80',
+  function () {
+    const rrn = '06020200080';
+    const formattedRrn = '06.02.02-000.80';
+    test(`${formattedRrn} is an invalid RRN, serial should be between 001 and 998`, function (assert) {
+      assert.false(isValidRijksregisternummer(rrn));
+    });
+  }
+);
```
JonasVanHoof, line 52: This test will fail but that is ok :)

This is fixed in https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/385

Rahien (approved):

PR #381, opened 2024-10-21
Lmb 922 fraction should only be required when burgemeest, CBS or GR

JonasVanHoof:
## Description

Fractie sleector is most of the time required. We only want it required when the bestuursorgaan is Gemeenteraad, College Burgemeest en College or Burgemeest (GR, CBS, Burgemeester )

## How to test

1. Go to a bestuursorgaan from the list where it is required
2. Create an eigen orgaan or go to an orgaan other than in the list, it should not be required
3. Create/edit mandataris forms use these (normal flow and IV)
4. Wijzig mandataris is also using this field but not the RDF one

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/254

### DIFF
--- a/app/components/mandatarissen/update-state.hbs
+++ b/app/components/mandatarissen/update-state.hbs
@@ -78,15 +78,13 @@
         </div>
       {{/if}}
       <div>
-        <AuLabel @required={{true}}>
-          Fractie
-        </AuLabel>
         {{#if this.load.isRunning}}
           <AuLabel for="input" class="skeleton-label au-u-margin-top" />
           <p class="skeleton-input-full au-u-margin-top au-u-margin-bottom"></p>
         {{else}}
           <Mandatarissen::FractieSelector
             @fractie={{this.selectedFractie}}
+            @isRequired={{this.isFractieSelectorRequired}}
             @onSelect={{this.updateFractie}}
             @allowClear={{false}}
             @bestuurseenheid={{this.bestuurseenheid}}
--- a/app/components/mandatarissen/update-state.hbs
+++ b/app/components/mandatarissen/update-state.hbs
@@ -86,7 +86,6 @@
             @fractie={{this.selectedFractie}}
             @isRequired={{this.isFractieSelectorRequired}}
             @onSelect={{this.updateFractie}}
-            @allowClear={{false}}
             @bestuurseenheid={{this.bestuurseenheid}}
             @bestuursperiode={{this.bestuursperiode}}
             @mandataris={{@mandataris}}
--- a/app/components/mandatarissen/update-state.js
+++ b/app/components/mandatarissen/update-state.js
@@ -16,6 +16,7 @@ import {
   notBurgemeesterStates,
 } from 'frontend-lmb/utils/well-known-uris';
 import { isDateInRange } from '../date-input';
+import { isRequiredForBestuursorgaan } from 'frontend-lmb/utils/is-fractie-selector-required';
 
 export default class MandatarissenUpdateState extends Component {
   @tracked newStatus = null;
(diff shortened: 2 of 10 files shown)


app/components/mandatarissen/update-state.hbs
```diff
@@ -78,15 +78,13 @@
...
-        </AuLabel>
         {{#if this.load.isRunning}}
           <AuLabel for="input" class="skeleton-label au-u-margin-top" />
           <p class="skeleton-input-full au-u-margin-top au-u-margin-bottom"></p>
         {{else}}
           <Mandatarissen::FractieSelector
             @fractie={{this.selectedFractie}}
+            @isRequired={{this.isFractieSelectorRequired}}
             @onSelect={{this.updateFractie}}
             @allowClear={{false}}
```
bfidlers, line 89: This allowClear is not used anymore.

app/components/rdf-input-fields/mandataris-fractie-selector.hbs
```diff
@@ -15,6 +15,7 @@
         @onClose={{fn (mut this.hasBeenFocused) true}}
         @person={{this.person}}
         @isInCreatingForm={{this.isPersonInForm}}
+        @allowClear={{not this.isRequiredForBestuursorgaan}}
```
bfidlers, line 18: Should this not be the `@isRequired`, the argument `@allowClear` is no longer used in the fractie-selector component.

app/components/mandatarissen/update-state.js
```diff
@@ -143,6 +151,7 @@ export default class MandatarissenUpdateState extends Component {
       !this.date ||
       this.inValidReplacement ||
       !this.hasChanges ||
+      (this.isFractieSelectorRequired && !this.selectedFractie) ||
```
bfidlers, line 154: Do we need this? Isn't this why we set the allowClear based on whether the fractie is required or not?

app/utils/form-context/application-context-meta-ttl.js
bfidlers, line 1: This feels way to complex, can't we just add a triple fractieRequired? Why do we need all this information in the applicationContext?

bfidlers (dismissed):
Needs some clean-up

app/components/rdf-input-fields/mandataris-fractie-selector.hbs
JonasVanHoof: It should be possible to allow clear when the fractie selector is not required IMO so I forgot to change the false to not isFractieSelector required (previous comment)

app/components/mandatarissen/update-state.js
JonasVanHoof: This is definitly needed as otherwise you can change the fractie in wijzig and you wont not able to save it

app/utils/form-context/application-context-meta-ttl.js
JonasVanHoof: You are suggesting to only set a triple is Fractie required in the application context? I think that is possible but thats only going to move this logic to the applicationContextMetaTtl file (where we build the application context)

```
const { isGR, isCBS, isBurgemeester } =
    getBestuursorgaanClassificaties(options);  // => this will become get is required from context and this value can directly be returned in this validation
  const requiredConditions = [isGR, isCBS, isBurgemeester];

  return requiredConditions.some((is) => is === true);
  ```

did you run the frontend of it?

app/components/mandatarissen/update-state.hbs
JonasVanHoof: Will update it so it reacts on the is fractie required

app/utils/form-context/application-context-meta-ttl.js
JonasVanHoof: Simplified it

Rahien (approved):

PR #382, opened 2024-10-22
allow generate rows for ocmw

Rahien:
## Description

image
image

## How to test
obvious

### DIFF
--- a/app/components/verkiezingen/prepare-legislatuur-section.hbs
+++ b/app/components/verkiezingen/prepare-legislatuur-section.hbs
@@ -6,16 +6,23 @@
     @bestuursorganen={{@bestuursorganen}}
   />
 {{else}}
-  <div class="au-u-flex au-u-flex--between au-u-margin-left au-u-margin-right">
-    <AuHeading @skin="3" class="au-u-flex-self-start">Samenstelling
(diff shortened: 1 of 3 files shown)


app/components/verkiezingen/prepare-legislatuur-section.hbs
```diff
@@ -6,52 +6,76 @@
...
+          }}
+            <div class="au-u-flex--inline">
+              <Shared::Tooltip
+                @showTooltip={{true}}
+                @tooltipText={{if
+                  (await @bestuursorgaan.isRMW)
+                  "Neem de mandatarissen over van de gemeenteraad"
+                  "Neem de mandatarissen over van het College van Burgemeester en Schepenen"
+                }}
+                @alignment="left"
```
bfidlers, line 55: I would align the tooltip right, since it is the rightmost button, and that way the alignment looks a bit better in my opinion.

bfidlers (approved):
Small comment, looks good otherwise.

app/components/verkiezingen/prepare-legislatuur-section.hbs
Rahien: alright!

PR #384, opened 2024-10-22
use lazy generators in our forms

Rahien:
## Description

This uses the experimental lazy generators that we added to the form helpers.
That way we reuse existing nodes when running the generators and we don't create zombie nodes.
## How to test

Go to the person form in the hidden forms route and edit a person. You will no longer see additional empty identifiers and birth date instances (zombies)


JonasVanHoof (approved):
Same changes i did when testing it out :D

PR #385, opened 2024-10-22
lmb-949: add extra check for serial part of rrn

JonasVanHoof:
## Description

We need to check if the serial part of rrn is between 001 and 998

## How to test

1. Fill in this rrn with serial part 000 and it should not be a valid rrn. `06.02.02-000.80`
2. When running the test suit of with flag `-f="RRN helpers" it should pass instead of fail now

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/379


JonasVanHoof:
We will not be doing this as this will lead us to far and we also do not want to be blocking input from the user e.g. a non belgian RRN.

PR #387, opened 2024-10-22
Ben/lmb 797 terminology consistency

bfidlers:
## Description

Update consistency of terminology.

## How to test

Look at the changes if they make sense.

## Links to other PR's

### DIFF
--- a/app/components/verkiezingen/kieslijst-splitter.hbs
+++ b/app/components/verkiezingen/kieslijst-splitter.hbs
@@ -158,16 +158,23 @@
       <div class="au-u-bold au-u-flex--inline">
         {{this.selectedKieslijst.lijstnaam}}
       </div>
-      in fracties
+      in
+      {{this.fractieText "de"}}
       {{#each this.selectedFracties as |fractie index|}}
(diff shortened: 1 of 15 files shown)


app/components/verkiezingen/kieslijst-splitter.hbs
```diff
@@ -158,16 +158,23 @@
       <div class="au-u-bold au-u-flex--inline">
         {{this.selectedKieslijst.lijstnaam}}
       </div>
-      in fracties
+      in
+      {{this.fractieText "de"}}
```
Rahien, line 162: didn't know you could already do this, feels more like something from polaris release. nice
but it means you need the action decorator right?

Rahien (approved):

app/components/verkiezingen/kieslijst-splitter.hbs
bfidlers: I think you need the action decorator indeed, not sure though, was more of an assumption.

bfidlers: yes you need it

PR #388, opened 2024-10-23
Lmb 796 export mandatarissen as csv

JonasVanHoof:
## Description

We want the mandatarissen to be downloadable. The file includes the required columns and the data will be filtered the way it was in the frontend. Creating this export is done in the mandataris service

## How to test

Open the mandataris search page or the organen>mandatarissen page and download the content. See if the results have the active filters applied. You will will also see previous mandatarissen of the person but this is how we want it.

## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/55

## Attachments

image


app/services/mandataris-api.js
```diff
@@ -104,4 +106,66 @@ export default class MandatarisApiService extends Service {
       include: 'fractietype',
     });
   }
+
+  async downloadAsCsv({
```
Rahien, line 110: why not just use an html-5 download link? this is the old way of doing things from before this existed but browser support seems good now

app/services/mandataris-api.js
```diff
@@ -104,4 +106,66 @@ export default class MandatarisApiService extends Service {
...
+    fractieIds = [],
+    hasFilterOnOnafhankelijkeFractie = false,
+    hasFilterOnNietBeschikbareFractie = false,
+    bestuursFunctieCodeIds = [],
+    sort = null,
+  }) {
+    const response = await fetch(
+      `${API.MANDATARIS_SERVICE}/mandatarissen/download`,
+      {
+        method: 'POST',
```
Rahien, line 124: feels weird that this is a post and not a get

app/services/mandataris-api.js
JonasVanHoof: Can you add a json body to this?

app/components/download-mandatarissen-from-table.hbs
```diff
@@ -0,0 +1,22 @@
+<Shared::Tooltip
+  @showTooltip={{true}}
+  @tooltipText="Exporteer mandatarissen naar CSV formaat"
+>
+  <AuLinkExternal
+    {{did-update
```
JonasVanHoof, line 6: Necessary to update the download link when a filter changes

app/services/mandataris-api.js
```diff
@@ -104,4 +104,30 @@ export default class MandatarisApiService extends Service {
...
+    bestuursorgaanId = null,
+    activeOnly = false,
+    persoonIds = null,
+    fractieIds = null,
+    hasFilterOnOnafhankelijkeFractie = false,
+    hasFilterOnNietBeschikbareFractie = false,
+    bestuursFunctieCodeIds = null,
+    sort = null,
+  }) {
+    const queryParams = [
```
JonasVanHoof, line 119: Rather do  it here than in the download component

app/components/download-mandatarissen-from-table.js
```diff
@@ -0,0 +1,89 @@
...
+      persoonIds: this.persoonIds,
+      fractieIds: this.fractieIds,
+      hasFilterOnOnafhankelijkeFractie: this.filterOnOnafhankelijkeFractie,
+      hasFilterOnNietBeschikbareFractie: this.filterOnNietBeschikbaarFractie,
+      bestuursFunctieCodeIds: this.bestuursFunctieCodeIds ?? [],
+      sort: this.args.sort,
+    });
+  });
+
+  async getBestuursorgaanInTijdForPeriod() {
```
JonasVanHoof, line 41: move one component up and pass as argument to the component

Rahien (approved):

PR #391, opened 2024-10-23
do not allow add/remove mandates at all

Rahien:
## Description

As the list of mandates that can be added is too weird they don't want users to be able to add/remove mandates to organs at all for now
## How to test
image


JonasVanHoof (approved):
Didn't see the buttons ;) Could have been a feature flag but we we can also see the history of the file

PR #394, opened 2024-10-25
do not allow just www valid uri checks

Rahien:
## Description

do not allow just www valid uri checks


JonasVanHoof (approved):
Looks good

PR #396, opened 2024-10-25
allow change of fraction in iv

Rahien:
## Description
allow change of fraction in iv
 or better, only force fraction limit in forms when creating outside of iv

image
image
image
image
image

## How to test

try updating fraction of people with and without fractions everywhere in the app

### DIFF
--- a/app/components/rdf-input-fields/mandataris-fractie-selector.js
+++ b/app/components/rdf-input-fields/mandataris-fractie-selector.js
@@ -42,7 +42,7 @@ export default class MandatarisFractieSelector extends InputFieldComponent {
   @tracked bestuursperiode;
   @tracked person;
   @tracked previousPerson;
-  @tracked isPersonInForm;
+  @tracked limitPersonFractionsToCurrent = false;
   @tracked isRequiredForBestuursorgaan;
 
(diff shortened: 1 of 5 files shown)


app/components/rdf-input-fields/mandataris-fractie-selector.js
```diff
@@ -68,13 +68,23 @@ export default class MandatarisFractieSelector extends InputFieldComponent {
...
       this.findPersonInForm.perform(),
       this.loadBestuursorganen(),
       this.loadProvidedValue(),
     ]);
     this.initialized = true;
   }
 
+  async checkIfShouldLimitFractions() {
+    this.limitPersonFractionsToCurrent = !!this.storeOptions.store.any(
+      new NamedNode('http://mu.semte.ch/vocabularies/ext/applicationContext'),
```
JonasVanHoof, line 81: Cant you just use EXT('applicationContext') ?

JonasVanHoof (approved):
Looks good,

image


Can change mandataaris outside of IV to only 2 options fractie one and onafhankelijk

In IV it is possible to change it to other fracties aswell, maybe a bit to loose these tables

PR #401, opened 2024-10-26
Ben/lmb 928 fractie selector burgemeester

bfidlers:
## Description

Extend burgemeester selector in the prepare legislature page. Now you can also select a fractie and beleidsdomeinen. There are still a few issues with it which need to be looked at, but they are not introduced by this change, see the issue section.

Screenshot 2024-10-26 at 11 08 14
Screenshot 2024-10-26 at 11 08 41

## How to test

Go to the prepare legislature page and add a burgemeester, this should also be added to the other mandates (voorzitter ...). The fractie and beleidsdomeinen should be added and it should also be possible to remove a burgemeester, which should result in the deletion of the voorzitter as well.

## Issues
There are a few issues that don't really have to do with this story, they already existed in the previous burgemeester selector or are an issue somewhere else.
- On the first load of the prepare-legislature page, the burgemeester mandate in CBS and the voorzitter in VB are not shown, but the are created behind the screens. This can be confusing.
- If you edit the aangewezen burgemeester in CBS, this also updates the burgemeester ambt, but it is only shown if you manually refresh
- In general the selector just feels too wonkey, with selecting the first one etc. We might need to improve this, maybe with linked mandatarissen or something like that, now it just does not feel very safe.
- Something weird happens with fetching beleidsdomeincodes, see comment in code, but depending on if you search for beleidsdomein-codes or concepts with the correct scheme, different results get shown. Probably an issue with the types that are set on these resources?

### DIFF
--- a/app/components/mandatarissen/beleidsdomein-selector-with-create.js
+++ b/app/components/mandatarissen/beleidsdomein-selector-with-create.js
@@ -67,7 +67,9 @@ export default class MandatenbeheerBeleidsdomeinSelectorWithCreateComponent exte
   @action
   select(beleidsdomeinen) {
     this.selected.setObjects(beleidsdomeinen);
-    this.args.onSelect(this.selected);
+    if (this.args.onSelect) {
+      this.args.onSelect(this.selected);
+    }
   }
 
   @action
--- a/app/components/mandatarissen/beleidsdomein-selector-with-create.js
+++ b/app/components/mandatarissen/beleidsdomein-selector-with-create.js
@@ -26,13 +26,35 @@ export default class MandatenbeheerBeleidsdomeinSelectorWithCreateComponent exte
   }
 
   async fetchOptions(searchData) {
-    let queryParams = {
+    // let queryParams = {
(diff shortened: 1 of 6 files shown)


app/components/mandatarissen/beleidsdomein-selector-with-create.js
```diff
@@ -26,13 +26,35 @@ export default class MandatenbeheerBeleidsdomeinSelectorWithCreateComponent exte
...
+        size: 9999,
+      },
     };
     if (searchData) {
-      queryParams['filter[label]'] = searchData;
+      queryParams['filter']['label'] = searchData;
     }
-    return await this.store.query('beleidsdomein-code', queryParams);
+    return await this.store.query('concept', queryParams);
   }
```
bfidlers, line 58: These two data fetches return different results, this really needs to be looked into. The one commented out seems the worst, so this can be removed. But first we need to know why.

Rahien (approved):
works as expected! created a ticket for the other issues you flagged

PR #402, opened 2024-10-26
make sure the correct bestuursperiode is used for selecting elected people

bfidlers:
## Description

Bestuursperiode and some other settings were not provided in the use of some person::Selector components. Which does not guarantee the selected people are elected in the correct bestuursperiode.

The commit b77e7f5d87cff243a0c0338eba916cb05512278a can be omitted if wanted, this just simplifies the arguments, since searchElected and allowCreate were just each others opposites.
 
## How to test

Double check each use of the person::Selector component and make sure you can only select elected people.


Rahien (approved):
It works, but it needs a migration because election results weren't exposed to ocmw. Not merging it right now as not urgent yet and i want the migration to be double checked

PR #407, opened 2024-10-29
 improve iv sorting

Rahien:
## Description

- Merge 'bcsd' service into installatievergadering. It was used for way more than just bcsd.
- track all mandataris instances in the installativergadering service and sort them
- use that sorting as a fallback if there is no rangorde to sort the mandataris instances of the other iv organs

image
(screenshot is of college vs vast bureau because in my testset that was the smallest, but gemeenteraad / rmw works the same way)
## How to test

- Go to the IV page and generate some data (or start from some test/anonymized production data)
- See that the order of the ocmw is dependent on the order of the gemeenteraad and similar for vast bureau vs college
- See that the order of the bcsd is dependent on the absolute rank of the person in other organs (mayor > schepen > voorzitter gr ... )
- See that if you change the order of e.g. gemeenteraadsleden, the order of ocmw changes immediately


JonasVanHoof (changes requested):
* What do you mean with OCMW here? "See that if you change the order of e.g. gemeenteraadsleden, the order of ocmw changes immediately"
* when changing the rangorde with the arrows the arrows are in reverse?  (how did this happen as you didn't touch the templates ;p)
* when you have rangorde 2,4,5 and you want to bring down mandataris from 5 to 3 this is not possible. Stuck on 4

* when changing the rangorde the table is direcly updated
* the voorzitter is always at the top of the table

JonasVanHoof (approved):

PR #408, opened 2024-10-31
fix binding einde custom org

Rahien:
## Description

fix binding einde custom org
## How to test

create a new custom org and check its end date

### DIFF
--- a/app/controllers/organen/index.js
+++ b/app/controllers/organen/index.js
@@ -105,7 +105,7 @@ export default class OrganenIndexController extends Controller {
           .toDate(),
       bindingEinde:
         similarBestuursorgaanInTijd?.bindingEinde &&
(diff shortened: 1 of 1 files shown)


app/controllers/organen/index.js
```diff
@@ -105,7 +105,7 @@ export default class OrganenIndexController extends Controller {
           .toDate(),
       bindingEinde:
         similarBestuursorgaanInTijd?.bindingEinde &&
-        moment(similarBestuursorgaanInTijd.bindingStart)
+        moment(similarBestuursorgaanInTijd.bindingEinde)
```
JonasVanHoof, line 108: Should we add a fix for the already created eigen bestuursorganen on prod?

First time tried it but it to the end date of my previous eigen orgaan
image

JonasVanHoof (approved):

PR #410, opened 2024-11-05
Ben/lmb 951 count nb persons

bfidlers:
## Description

In installatievergadering count the number of persons with mandates, not the number of mandatarissen.
Screenshot 2024-11-05 at 09 27 48

## How to test

Go to the installatievergadering, add some mandatarissen and check if the counts are correct.


JonasVanHoof (approved):
Good fix
image

PR #411, opened 2024-11-05
only save an independent fractie if it is going to be used

Rahien:
## Description

only save an independent fractie if it is going to be used
## How to test

go to iv, notice no independent fraction is created when creating mandataris
go to non-iv mandataris, correct mistakes, see that the independent fraction is only created on select
go to non-iv mandataris, see that update state still works as before
go to a mandataris and check if the 'maak onafhankelijk' feature still works


bfidlers (approved):
Much better, the fractie-selector might require another look, because it is weird to begin with that the onafhankelijke fractie is created in iv. But an issue for later.

PR #414, opened 2024-11-06
Ben/lmb 1019 warning no chair

bfidlers:
## Description

Give a warning if no voorzitter is selected for an organ in the preperation of legislature. The warning only shows if at least one other mandataris has been added. It does also show if a voorzitter is added without a person (VB), but not if that is the only mandataris added (like in the initial situation).

Screenshot 2024-11-06 at 17 23 23

## How to test

Check if no warnings are given in the initial state.
Check if warnings are given if you add a lid, but not a voorzitter.
Check if the warning disappears if you add a voorzitter.


JonasVanHoof (approved):
Looks good

PR #418, opened 2024-11-08
lmb-934: date input can handle undefined start date for bindingStart

JonasVanHoof:
## Description

The date input field cannot handle returns invalid in the form when no min date is set.

## How to test

Create a mandataris in an eigen orgaan in the politieraad and it should create the mandataris even so the binding start and end are not defined.

### DIFF
--- a/app/utils/form-validations/mandataris-date-in-bestuursperiod.js
+++ b/app/utils/form-validations/mandataris-date-in-bestuursperiod.js
@@ -11,10 +11,14 @@ export const isValidMandatarisDate = ([dateLiteral], options) => {
   const date = new Date(dateLiteral.value);
   const period = loadBestuursorgaanPeriodFromContext(options);
   let maxDate = period.endDate;
+  let startDate = period.startDate;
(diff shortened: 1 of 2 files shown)


app/utils/form-validations/mandataris-date-in-bestuursperiod.js
```diff
@@ -11,10 +11,14 @@ export const isValidMandatarisDate = ([dateLiteral], options) => {
   const date = new Date(dateLiteral.value);
   const period = loadBestuursorgaanPeriodFromContext(options);
   let maxDate = period.endDate;
+  let startDate = period.startDate;
 
   if (moment(period.endDate).isSame(moment(NULL_DATE))) {
     maxDate = null;
   }
+  if (moment(period.startDate).isSame(moment(NULL_DATE))) {
```
JonasVanHoof, line 19: Should we default this to the januari the first of the bestuursperiode year?

Rahien (approved):

PR #420, opened 2024-11-12
Ben/lmb 1036 show election results faciliteitengemeentes

bfidlers:
## Description

For faciliteitengemeentes, the prepare iv can have multiple verkiezingsresultaten. If the iv has one type of verkiezingsresultaat (e.g. gemeenteraadsverkiezingen) you will go to that url immediately, if there are more, you have to select which one you want to go to.

Screenshot 2024-11-12 at 14 10 17

Not that happy with the UI, but don't care enough to waste too much time on it.

I made a mapping of kandidatenlijsten to types of verkiezingen, which is probably a bit too extensive (we only need ocmw and gemeente), but I would rather match all than miss something.

## How to test

Go to a normal gemeente, check if you go to the results of the elections immediately. Go to a faciliteitsgemeente (e.g. Drogenbos) and check if you get the modal and it works.

### DIFF
--- a/app/templates/verkiezingen/installatievergadering.hbs
+++ b/app/templates/verkiezingen/installatievergadering.hbs
@@ -135,7 +135,8 @@
               @route="verkiezingen.verkiezingsuitslag"
               @model={{verkiezing.id}}
               role="menuitem"
-            >{{verkiezing.getLijsttype}}
+            >Verkiezingsuitslag
+              {{verkiezing.getType}}
             </AuLink>
           {{/each}}
         </AuDropdown>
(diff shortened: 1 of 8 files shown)


app/templates/verkiezingen/installatievergadering.hbs
```diff
@@ -106,6 +113,37 @@
...
+      Selecteer welke uitslag je wilt bekijken.
+      <li class="au-c-list-horizontal__item au-u-margin-top-small">
+        <AuDropdown
+          @title="Selecteer verkiezingsuitslag"
+          @alignment="left"
+          @skin="secondary"
+          role="menu"
+        >
+          {{#each @model.verkiezingen as |verkiezing|}}
+            <AuLink
```
Rahien, line 134: not super annoying but this modal stays open if you click back in your browser navigation. maybe you can use setupcontroller to set this.verkiezingsUitslagModal to false? If not, don't care

Rahien (approved):

PR #421, opened 2024-11-12
Ben/lmb 1037 hide duplicate kandidatenlijsten

bfidlers:
## Description

In faciliteitsgemeentes, kandidatenlijsten were shown for gemeente and ocmw. Hide those for the OCMW, we don't really care about them.

## How to test

Go to faciliteitsgemeente, check if kandidatenlijsten are only shown once and check if the flow of iv and syncing still works as expected.


bfidlers:
Instead of the promise hell, I can also just filter kandidatenlijsten with type KANDIDATENLIJST_GEMEENTE, but this feels more correct.

bfidlers:
> Instead of the promise hell, I can also just filter kandidatenlijsten with type KANDIDATENLIJST_GEMEENTE, but this feels more correct.

Apparently the promise hell was entirely circumventible

Rahien (approved):

PR #423, opened 2024-11-18
Ben/lmb 952 improve become onafhankelijk

bfidlers:
## Description

This PR does a bit more than asked.
It does make it possible for people to become onafhankelijk in the mandaten page of that person, even if they don't have a fractie.
But it also fixes some issues with the become onafhankelijk. Before it created an onafhankelijke fractie per bestuursorgaan in de tijd, now it checks the bestuursperiode, which also simplifies some code.

## How to test

Check if you can become onafhankelijk in the mandaten page of a person, even if they don't have a fractie.
Also check it is the same fractie if they have multiple mandates in the same bestuursperiode of which one is already onafhankelijk.


Rahien (approved):
works!

PR #425, opened 2024-11-18
don't await restartable tasks

Rahien:
## Description

see https://ember-concurrency.com/docs/task-cancelation-help
## How to test

test the person selector in the forms


bfidlers (approved):
Person Selector still works in prepare installatievergadering for me

PR #427, opened 2024-11-19
Ben/lmb 1052 create people ocmw

bfidlers:
## Description

If people get created that belong to an orgaan that does not really belong in the bestuurseenheid of that graph, the person information is copied to the correct graph. 
An example is bcsd, if a person is created in the installatievergadering for the bcsd, this person is also copied to the ocmw.

## How to test

Two things should be tested:
- Create a person which should not be copied (e.g. in a custom orgaan such as Jeugdraad), and check that the person is not copied to the OCMW or any other graph for that matter.
- Create a person in the bcsd in the iv and check that this person is copied correctly to the ocmw. His mandate shouldn't be there yet, but the person data should.

## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/67

## Remarks
This copy function (`movePersons`) also exists in the move-ocmw-organs call, which is called whenever the iv is concluded. I left it there for completeness and because it can be gemeente people should be copied which didn't exist in the ocmw yet (although, this is all election data, which should be in both). But this shouldn't result in any issues as this data shouldn't be copied since it already exists.

### DIFF
--- a/app/components/person/selector.js
+++ b/app/components/person/selector.js
@@ -50,7 +50,7 @@ export default class PersonSelectorComponent extends Component {
     this.closeModal();
     this.persoonApi.putPersonInRightGraph(
       instanceId,
-      this.args.bestuursorgaanIT
+      this.args.bestuursorgaanIT.id
     );
     this.args.onUpdate(this.person);
   }
--- a/app/components/person/selector.js
+++ b/app/components/person/selector.js
@@ -48,11 +48,11 @@ export default class PersonSelectorComponent extends Component {
     const persoon = await this.store.findRecord('persoon', instanceId);
     this.person = persoon;
     this.closeModal();
-    this.persoonApi.putPersonInRightGraph(
+    await this.persoonApi.putPersonInRightGraph(
       instanceId,
       this.args.bestuursorgaanIT.id
     );
-    this.args.onUpdate(this.person);
+    await this.args.onUpdate(this.person);
   }
 
   @action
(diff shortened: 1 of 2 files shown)


app/components/person/selector.js
```diff
@@ -47,6 +48,10 @@ export default class PersonSelectorComponent extends Component {
     const persoon = await this.store.findRecord('persoon', instanceId);
     this.person = persoon;
     this.closeModal();
+    this.persoonApi.putPersonInRightGraph(
```
Rahien, line 51: this isn't awaited so if it the other statement continues (and it does so before this method terminates)
I would wait for it and not execute the update if it fails

also: i see the onupdate isn't awaited, it would be nice to do that because this is called in the oncreate of the form instance so that button will stay in the loading state while all this is going on

app/services/persoon-api.js
```diff
@@ -57,4 +57,22 @@ export default class PersoonApiService extends Service {
...
+      `${API.MANDATARIS_SERVICE}/personen/${persoonId}/put-person-in-right-graph/${bestuursorgaanID}`,
+      {
+        method: 'POST',
+      }
+    );
+    const jsonReponse = await response.json();
+
+    if (response.status !== STATUS_CODE.OK) {
+      console.error(jsonReponse.message);
+      throw {
```
Rahien, line 72: I see this is done in other places too but i believe it's actually not good practice and you should throw instances of an error object. the benefit is that you get a stack trace if it ends up with the user (or in our glitchtip environment).
don't necessarily do it for the existing cases, feel free to make a ticket for them

Rahien (changes requested):

Rahien (approved):

PR #441, opened 2024-12-09
Ben/lmb 1153 improve burgemeester selector

bfidlers:
## Description

Update burgemeester selector to be more user friendly. Three things should have happened:
- You can now edit the burgemeester in the prepare legislature page
- You can now select a start date in the modal when selecting a burgemeester
- In the tables the startdate is know shown as well.

## How to test

Go to the prepare legislature page and play around with the burgemeester selector.

### DIFF
--- a/app/components/mandaat/burgemeester-selector.js
+++ b/app/components/mandaat/burgemeester-selector.js
@@ -17,6 +17,7 @@ export default class MandaatBurgemeesterSelectorComponent extends Component {
   @service store;
   @service installatievergadering;
   @service('mandataris') mandatarisService;
+  @service fractieApi;
 
   @tracked persoon = null;
   @tracked aangewezenBurgemeesters;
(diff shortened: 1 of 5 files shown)


app/components/mandaat/burgemeester-selector.js
```diff
@@ -196,4 +210,19 @@ export default class MandaatBurgemeesterSelectorComponent extends Component {
...
+    this.isEditFormInitialized = false;
+  }
+
+  @action
+  async saveBurgemeesterChanges({ instanceId }) {
+    await this.fractieApi.updateCurrentFractie(instanceId);
+    this.closeEditBurgemeesterModal();
+    if (this.args.onUpdateBurgemeester) {
+      setTimeout(() => this.args.onUpdateBurgemeester(), 1000);
+    }
```
bfidlers, line 226: This should trigger a router refresh, but it does not seem to be happening, anybody got an idea why?

JonasVanHoof: Added a comment in the RC chat for this for using the draft mandataris component togheter with the IV service

PR #442, opened 2024-12-09
Lmb 941 | add warning if bcsd voorzitter is not schepen

JonasVanHoof:
## Description

When the selected voorzitter in BCSD does not have a mandaat schepen/toegevoegde schepen than we have to show a warning for it.

extra: As there already is a bcsdVoorzitter alert we will extends this one

## How to test

1. Add a bcsd voorzitter without a person => no alert
2. Assign a person that is not a schepen already => alert should be shown
3. Add a schepen mandaat for the voorzitter persoon => alert should be gone
4. Remove the schepen mandaat => alert should pop up again

## Attachments

image


app/components/verkiezingen/bcsd-voorzitter-alert.hbs
```diff
@@ -1,9 +1,13 @@
...
+{{#if (await @bestuursorgaanInTijd.isBCSD)}}
+  <div {{did-update this.mandatarissenUpdated @trigger}}>
+    {{#if this.messages}}
+      <div class="au-o-box">
+        <AuAlert @skin="warning" @icon="alert-triangle" @closable={{true}}>
+          {{#each this.messages as |warning|}}
+            <p> - {{warning.message}}</p>
+          {{/each}}
+        </AuAlert>
+      </div>
```
JonasVanHoof, line 10: image

app/components/verkiezingen/bcsd-voorzitter-alert.js
```diff
@@ -1,130 +1,99 @@
...
-      this.bestuursorganenInTijd,
-      async (bestInTijd) => {
-        return await hasClassificationFunc(bestInTijd);
+    if (hasVoorzitter) {
+      const isLidRMWOrVastBureau =
+        await this.findMandatarisForOneOfBestuursfunctieCodes(
+          hasVoorzitter.persoon,
+          [MANDAAT_LID_RMW_CODE_ID, MANDAAT_LID_VAST_BUREAU_CODE_ID]
+        );
+      if (!isLidRMWOrVastBureau) {
```
Rahien, line 64: so this warning can be removed because it is included in the first warning about them having to be schepen

Rahien (changes requested):

app/components/verkiezingen/bcsd-voorzitter-alert.js
JonasVanHoof: For me its a different situation  but if you say it's like that ok 😰

app/components/verkiezingen/bcsd-voorzitter-alert.hbs
```diff
@@ -1,9 +1,13 @@
...
-  {{/if}}
-</div>
\ No newline at end of file
+{{#if (await @bestuursorgaanInTijd.isBCSD)}}
+  <div {{did-update this.mandatarissenUpdated @trigger}}>
+    {{#if this.messages}}
+      <div class="au-o-box">
+        <AuAlert @skin="warning" @icon="alert-triangle" @closable={{true}}>
+          {{#each this.messages as |warning|}}
+            <p> - {{warning.message}}</p>
```
Rahien, line 7: would remove the -, it looks silly when there is only one warning and if there are multiple, it doesn't hurt that there isn't any - in front of them

app/components/verkiezingen/bcsd-voorzitter-alert.js
```diff
@@ -1,130 +1,75 @@
...
+        };
+      })
+    );
+    const hasVoorzitter = mapping.find(
+      (mapping) => mapping.persoon && mapping.isVoorzitter
+    );
+    if (hasVoorzitter) {
+      const schepen = await this.findMandatarisForOneOfBestuursfunctieCodes(
+        hasVoorzitter.persoon,
+        [MANDAAT_SCHEPEN_CODE_ID, MANDAAT_TOEGEVOEGDE_SCHEPEN_CODE_ID]
```
Rahien, line 38: i think burgemeester and aangewezen burgemeester should also be allowed because they just have to be member of vast bureau so the mayor is allowed to do this himself i think

Rahien: indeed: chair of vb (so the mayor) can do this too: https://www.vlaanderen.be/lokaal-bestuur/organisatie-en-werking/bijzonder-comite-voor-de-sociale-dienst

JonasVanHoof: Adde burgemeester and aangewezen to the mandaten array , https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/442/commits/f7d945e5a089ce26f3ded2bc1ceae994b244e057

Rahien (approved):

PR #446, opened 2024-12-13
Lmb 858 ember context

JonasVanHoof:
## Description

Use ember context for showing multiple alerts components

## How to test

/


JonasVanHoof:
This logic for alert group was postponed

PR #448, opened 2024-12-17
Lmb 1192 remove active toggle for past organen

JonasVanHoof:
## Description

Do not show the toggle when there cannot be active  mandatrissen for that bestuursperiod. 

## How to test

in 2012 -2019 the toggle should be hidden.

### DIFF
--- a/app/controllers/organen/orgaan/mandatarissen.js
+++ b/app/controllers/organen/orgaan/mandatarissen.js
@@ -22,7 +22,6 @@ export default class OrganenMandatarissenController extends Controller {
   @tracked isCreatingMandataris = false;
   @tracked createdMandataris = false;
   @tracked activeOnly = true;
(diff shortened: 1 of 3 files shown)


app/controllers/organen/orgaan/mandatarissen.js
```diff
@@ -77,4 +78,25 @@ export default class OrganenMandatarissenController extends Controller {
   toggleActiveOnly() {
     this.activeOnly = !this.activeOnly;
   }
+
+  @action
+  async isSelectedPeriodCurrent() {
```
Rahien, line 83: shouldn't this just be a getter?

JonasVanHoof: Hm but than I cannot set the tracked properties for it

Rahien (approved):

PR #450, opened 2024-12-19
Lmb 892 | different home page for admin

JonasVanHoof:
## Description

When logged in as an administrator you see the same modules as you are logged in as a normal bestuurseenheid. This would be okey if you could click on those modules without getting an error.

Show a different home page for a user that is logged in as an administrator.

## How to test

- Log in as a bestuurseenheid => everything works as before
- Log in as an administrator
  -  you should see a different home page
  - you can directly simulate another bestuurseenheid 
  - you can set a system message for all bestuurseenheden

## Attachments

**Logged in as an admin**
image

**As an admin that simulates another bestuurseenheid**
image

**As normal bestuurseenheid**
image


app/components/impersonation-menu.hbs
```diff
@@ -1,7 +1,7 @@
 <div class="au-u-flex au-u-flex--vertical-center">
   <AuPill @icon="user">{{this.adminLabel}}</AuPill>
 
-  <AuDropdown
```
Rahien, line 4: remove instead of comment

app/components/shared/main-menu.js
```diff
@@ -5,8 +5,21 @@ import { service } from '@ember/service';
 export default class SharedMainMenuComponent extends Component {
   @service currentSession;
   @service features;
+  @service impersonation;
 
   get showLegislatuurModule() {
     return this.currentSession.showLegislatuurModule;
   }
+
+  get areModulesShown() {
```
Rahien, line 14: why have this and now just showCommonModules?

app/components/shared/main-menu.hbs
```diff
@@ -1,43 +1,39 @@
...
-  </ul>
-</nav>
\ No newline at end of file
+{{#if this.areModulesShown}}
+  <div class="au-c-sidebar__header">
+    <AuHeading @level="6" @skin="6">Menu modules</AuHeading>
+  </div>
+  <nav>
+    <ul class="au-c-list-navigation">
+      {{#if this.showCommonModules}}
```
Rahien, line 7: this if seems redundant now since areModulesShown is the same as showCommonModules?

app/controllers/application.js
```diff
@@ -16,8 +16,13 @@ export default class ApplicationController extends Controller {
 
   appTitle = 'Lokaal mandatenbeheer';
 
-  get isIndex() {
-    return this.router.currentRouteName === 'index';
+  get isNavigationVisible() {
```
Rahien, line 19: should this not live in the navigation bar component instead of passing it down as a prop?

app/controllers/overzicht/admin.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default class OverzichtAdminController extends Controller {
+  @service currentSession;
+  @service impersonation;
+
+  get isImpersonating() {
+    return this.impersonation.isImpersonating;
+  }
+
+  get postfixBestuurseenheidLabel() {
```
Rahien, line 14: this is a prefix apparently

app/routes/admin-panel/global-system-message.js
```diff
@@ -7,12 +7,15 @@ export default class AdminPanelGlobalSystemMessageRoute extends Route {
   @service session;
   @service router;
   @service('globalSystemMessage') messageService;
+  @service impersonation;
 
   beforeModel(transition) {
     this.session.requireAuthentication(transition, 'login');
-
-    if (!this.currentSession.showAdminFeatures) {
```
Rahien, line 14: i sortof liked having this as a simple getter. don't we want to keep this?

app/routes/admin-panel/index.js
```diff
@@ -6,15 +6,15 @@ export default class AdminPanelRoute extends Route {
...
   @service session;
   @service router;
+  @service impersonation;
 
   beforeModel(transition) {
     this.session.requireAuthentication(transition, 'login');
-
-    if (!this.currentSession.showAdminFeatures) {
-      this.router.replaceWith('index');
+    if (
```
Rahien, line 13: because now we are repeating ourselves and if we want to change this it will be in many places

app/routes/overzicht.js
```diff
@@ -0,0 +1,16 @@
...
+
+export default class OverzichtRoute extends Route {
+  @service currentSession;
+  @service session;
+  @service router;
+
+  beforeModel(transition) {
+    this.session.requireAuthentication(transition, 'login');
+    if (this.currentSession.isAdmin) {
+      this.router.replaceWith('overzicht.admin');
```
Rahien, line 13: odd. i would just have kept one route and would have made it conditional with {{#if }} statements, no?

app/services/current-session.js
```diff
@@ -130,8 +129,4 @@ export default class CurrentSessionService extends Service {
     }
     return roles.includes(ADMIN_ROLE);
   }
-
-  get showAdminFeatures() {
```
Rahien, line 134: so this one i would keep (but maybe modify to what you needed) as said before

app/templates/application.hbs
```diff
@@ -46,11 +46,18 @@
             Wissel van bestuurseenheid
           </AuLink>
 
-          {{#unless this.currentSession.showAdminFeatures}}
+          {{#if
+            (not
```
Rahien, line 50: again another place where you repeat the same logic, but now in a different way of writing it down. If we ever update this we will never find this. please keep the getter

app/templates/application.hbs
```diff
@@ -81,45 +88,8 @@
       </li>
     {{/if}}
   </AuMainHeader>
-  {{#unless this.isIndex}}
-    <Shared::NavigationBar />
-  {{/unless}}
-  {{#if this.session.isAuthenticated}}
-    <AuMainContainer as |main|>
```
Rahien, line 88: doesn't this mean the maincontainer is gone in all other routes than the overzicht route? please keep it around on the application level. It's there so appuniversum can put general app stuff in this component

app/templates/overzicht/admin.hbs
```diff
@@ -0,0 +1,95 @@
...
+              voor de gewone bestuurseenheden zichtbaar zijn, een overzicht vind
+              je hieronder.
+            </p>
+          </AuContent>
+          <AuHeading @level="2" @skin="3" class="au-u-margin-bottom">Beschikbare
+            modules</AuHeading>
+          <ul class="au-o-grid au-o-grid--small">
+            <li class="au-o-grid__item au-u-1-2@medium au-u-1-3@large">
+              <LoketModuleCard @icon="settings">
+                <:title>Login als een bestuurseenheid</:title>
```
Rahien, line 44: this is way better than the little icon in the top :+1:

Rahien (changes requested):

app/routes/overzicht.js
JonasVanHoof: Isn't the index component not getting to big in that case?

app/controllers/application.js
JonasVanHoof: Im not always sure what the most explicit is, lmiting something in the component or seeing that it is limited when looking where the component is implemented.

Will move it in the component

app/components/shared/main-menu.js
JonasVanHoof: Yeah sometimes I think what if there are more.. but thats not the case atm so ill remove it

app/components/impersonation-menu.hbs
JonasVanHoof: Wasn't sure if you where a fan of the new one but apparently you are! Will remove. Should i also move the admin label to our index js file? Than this component can be removed

JonasVanHoof: Removed component

app/controllers/application.js
JonasVanHoof: Moved it to the NavigationBar component

app/routes/admin-panel/global-system-message.js
JonasVanHoof: Indeed alot easier in use

app/templates/application.hbs
JonasVanHoof: Added it again

app/components/shared/main-menu.js
```diff
@@ -5,8 +5,17 @@ import { service } from '@ember/service';
...
     return this.currentSession.showLegislatuurModule;
   }
+
+  get showCommonModules() {
+    if (!this.currentSession.isAdmin) {
+      return true;
+    }
+
+    return this.impersonation.isImpersonating;
+  }
```
JonasVanHoof, line 20: Replaced this also with the getter from the currentSession service

Rahien (approved):

PR #451, opened 2024-12-20
Lmb 414 | split group mandates in person table

JonasVanHoof:
## Description

We do not want to show the mandates of a person as a comma separated link string. Make it possible to show a dropdown of the rows for the mandaat with the orgaan.

## How to test

Go to mandatarissen module and click open a persoon. The mandaat and orgaan should be clickable and bring you to the correct place.


app/components/mandatarissen/persoon-table-row.js
```diff
@@ -0,0 +1,75 @@
...
+  @tracked isSubRowOpen;
+  @tracked subRows = A();
+
+  @action
+  openCloseSubRows() {
+    this.isSubRowOpen = !this.isSubRowOpen;
+  }
+
+  @action
+  async getSubRowData() {
```
JonasVanHoof, line 22: Not yet convinced to do this in the route

app/components/mandatarissen/persoon-table-row.js
```diff
@@ -0,0 +1,75 @@
...
+
+import { tracked } from '@glimmer/tracking';
+import { A } from '@ember/array';
+import { action } from '@ember/object';
+import { service } from '@ember/service';
+
+import { foldMandatarisses } from 'frontend-lmb/utils/fold-mandatarisses';
+
+export default class MandatarissenPersoonTableRowComponent extends Component {
+  @service store;
```
JonasVanHoof, line 11: remove this, unused

app/components/mandatarissen/persoon-table-row.hbs
```diff
@@ -0,0 +1,100 @@
...
+    class="au-c-link au-u-medium"
+  >
+    {{@persoon.achternaam}}
+  </LinkTo>
+</td>
+
+{{! Fractie}}
+<td>
+  {{#unless this.isSubRowOpen}}
+    <Mandaat::FoldedFracties
```
JonasVanHoof, line 35: TODO: Do this inline

JonasVanHoof: To much logic in the component, added an arugment for the component to use the provided mandatarissen

app/components/mandatarissen/persoon-table-row.js
JonasVanHoof: Reason: this component should display subrows, the data structure and template are only used here. If we move this data structure to the route all other components do not have any connecting with it.

app/components/mandatarissen/persoon-table-row.hbs
```diff
@@ -0,0 +1,100 @@
+{{! Naam}}
+<td {{did-insert this.getSubRowData}}>
+  <div class="au-u-flex au-u-flex--row">
+    <span style={{if (eq this.subRows.length 1) "visibility: hidden" ""}}>
```
Rahien, line 4: this is not enough. you need a getter to check if it has either more than 1 subrow OR if the only subrow is a mandataris in eigen orgaan

Rahien (changes requested):

app/components/mandatarissen/bestuursorganen-as-links.hbs
```diff
@@ -0,0 +1,12 @@
+<div {{did-insert (perform this.setup)}}>
```
Rahien, line 1: i don't like it but i tried to get rid of it and it's not straight forward

app/components/mandatarissen/persoon-table-row.js
Rahien: BUT fetching data in ember should be the responsibility of the routes (except for certain edge cases where that's not possible). In this case that's very possible AND it makes it so that we don't have to do a did-insert which we should avoid

app/components/mandatarissen/persoon-table.js
```diff
@@ -0,0 +1,14 @@
+import Component from '@glimmer/component';
+
+import { action } from '@ember/object';
+
+export default class MandatarissenPersoonTableComponent extends Component {
+  @action
+  mandatarissenFor(persoon) {
```
Rahien, line 7: should be done in the route. This is a hack, not ember

app/components/mandatarissen/persoon-table-row.js
JonasVanHoof: Moved it

app/components/mandatarissen/persoon-table.js
JonasVanHoof: Removed file

app/components/mandatarissen/persoon-table-row.hbs
JonasVanHoof: removed did-insert + changed subRows.length to mandatarissen.length (see new logic)

Rahien (approved):
i think we are still doing too many calls, but we've always done that and fixing it now would be premature optimization.

PR #456, opened 2025-01-09
clean up session more explicitly when ending up on the session expired page

Rahien:
## Description

When users end up here, log them out and redirect them to login when clicking the button. Else ACM/IDM remembers their session and they end up looping.

## How to test

Hard to tests without acm/idm. Try it in the test environment or by overriding your /etc/hosts


bfidlers (approved):
The redirection is a bit jittery.
But it works as expected!

PR #455, opened 2025-01-09
Lmb 182  | can modify custom field

JonasVanHoof:
## Description

We want the user to edit the custom form field. Adding an edit icon that opens a modal with the editable fields.

## How to test

Edit a custom field in the mandataris extra info card

## Links to other PR's

- App started from https://github.com/lblod/app-lokaal-mandatenbeheer/pull/346
- base of this PR is not master ⚠️ 
- Form content https://github.com/lblod/form-content-service/pull/69

PR #457, opened 2025-01-10
allow changing field order

Rahien:
## Description

Allow chaning field order. Up and down is done with ugly buttons. I think we need a hamburger menu that includes up, down, edit, remove. But we can do that later.

image

added a cancel button to the modal because i kept clicking 'verwijder' to close

## How to test

Move a field up and down all the way and across existing (unmodifiable) form fields
## Links to other PR's

- https://github.com/lblod/form-content-service/pull/66


JonasVanHoof (commented):
Just added the buttons to the modify forms branch
https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/455

PR #458, opened 2025-01-13
Lmb 1226 | can mark custom field required

JonasVanHoof:
## Description

You should be able to add a telephone number field that has phone validation + make it possible to mark a field as required

## How to test

Add these fields to a form and see if the validations are correct.

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/349
- https://github.com/lblod/form-content-service/pull/68

## Attachments

image


app/components/rdf-input-fields/crud-custom-field-modal.js
```diff
@@ -251,4 +256,8 @@ export default class RdfInputFieldCrudCustomFieldModalComponent extends Componen
 
     return 'Pas een veld aan';
   }
+
+  get canShowRequiredToggle() {
+    return this.displayType.uri !== ADRES_CUSTOM_DISPLAY_TYPE;
+  }
```
JonasVanHoof, line 262: Important the adress component cannot handle the validations as other rdf inputs do

Rahien: should we not just make it handle that then? Its value is just a uri so i think the default checks should work for it?

Rahien (changes requested):
apart from this, when changing field type a couple of times, the "dit veld is verplicht" error message is duplicated and see if you can fix the behavior when something that is not a number or date is filled in but the type is changed to number or date. Nothing is shown and the form can be saved because there is a value. Maybe it should clear that value in that case?

app/components/rdf-input-fields/crud-custom-field-modal.js
JonasVanHoof: It works with an adres as displaytype but I can't seem to make it work for the library entry "Verblijfplaats", the path of the validation and the field are the same. I removed the restriction of the required toggle not showing, the remark will than be that the verblijfplaats field will CAN be filled in as empty.

Rahien (approved):

PR #459, opened 2025-01-14
don't allow changing bekrachtigd status in bulk view

Rahien:
## Description

Don't allow changing bekrachtigd status in bulk view

image
image
image

## How to test

Try all in your power to change the bekrachtigd status (queries is cheating)


bfidlers (approved):
Not sure if it is super clear for users, why they can't select mandatarissen with status bekrachtigd, and probably the warning in the modal needs to be updated a bit with this change, but otherwise it works!

bfidlers (approved):
Rereviewed, looks clearer like this.

PR #462, opened 2025-01-16
min max mandatarissen

Rahien:
## Description

Warning for min/max mandatarissen

image
image
image
image

## How to test

Add too many/few mandatarissen in an orgaan. you can use verhinderd to easily toggle mandatarissen on/off. Burgemeesters are not checked at the moment but the code is there for when they would be.

### DIFF
--- a/app/components/min-max-mandataris-warning.js
+++ b/app/components/min-max-mandataris-warning.js
@@ -1,0 +1,129 @@
+import Component from '@glimmer/component';
+import moment from 'moment';
+import { service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { restartableTask } from 'ember-concurrency';
+import {
+  MANDAAT_AANGEWEZEN_BURGEMEESTER_CODE,
+  MANDAAT_BURGEMEESTER_CODE,
+  MANDATARIS_VERHINDERD_STATE,
+} from 'frontend-lmb/utils/well-known-uris';
+
+export default class MinMaxMandatarisWarningComponent extends Component {
+  @service store;
+  @tracked warningMessages;
+  @tracked currentIndex = 0;
+
+  constructor() {
+    super(...arguments);
+    this.setup.perform();
+  }
+
+  get dateToCheckFor() {
+    const now = moment();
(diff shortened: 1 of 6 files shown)


app/controllers/organen/orgaan/mandatarissen.js
```diff
@@ -86,12 +86,20 @@ export default class OrganenMandatarissenController extends Controller {
...
     }
 
     return isCurrent;
   }
+
+  get personCount() {
+    const personIds = this.model.mandatarissen.map(
+      (mandataris) => mandataris.mandataris.isBestuurlijkeAliasVan?.id
+    );
+    return new Set(personIds).size;
```
bfidlers, line 103: Not sure about this... We count all people here, also people who are verhinderd or are not active at that time. But this might be confusing when it says e.g. There are 50 people, but there might be a warning that there are not enough mandatarissen (if the limit is e.g 47)...

app/components/min-max-mandataris-warning.js
```diff
@@ -0,0 +1,129 @@
...
+        mandataris.get('bekleedt.bestuursfunctie.uri') ===
+        MANDAAT_AANGEWEZEN_BURGEMEESTER_CODE
+      ) {
+        targetMandaat = burgemeesterMandaat.id;
+      }
+      const start = moment(mandataris.start);
+      const einde = moment(mandataris.einde || '3000-01-01');
+      if (
+        date.isBetween(start, einde, 'day', '[]') &&
+        mandataris.get('status.uri') !== MANDATARIS_VERHINDERD_STATE
```
bfidlers, line 75: Just to double check, is it right waarnemend should be included?

bfidlers (approved):
Looks good, just some small remarks / questions in the code.

app/controllers/organen/orgaan/mandatarissen.js
Rahien: probably best to just remove the count then?

bfidlers: I think it might be beneficial, would revisit it if we get some negative feedback maybe?

PR #463, opened 2025-01-16
Lmb 1222 | read only custom form view

JonasVanHoof:
## Description

We want an experimental read-only form view for custom forms

## How to test

Look at the mandataris extra info card in the frontend and see if it reflects with what is in the form

## Attachments

**V1**
image

**V2**
> Form
image

> Read-only
image

### DIFF
--- a/app/components/editable-form.js
+++ b/app/components/editable-form.js
@@ -1,6 +1,5 @@
 import Component from '@glimmer/component';
 
-import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { service } from '@ember/service';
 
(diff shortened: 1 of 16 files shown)


app/components/editable-form.js
```diff
@@ -48,4 +48,9 @@ export default class EditableFormComponent extends Component {
   get formDefinition() {
     return this.currentForm;
   }
+
+  @provide('is-read-only')
```
Rahien, line 52: we're now providing 3 contexts. But really this should be one context right? 
```
@provide('form-context)
get formContext(){
  return {
      onFormUpdate: ...
      formDefinition: ...
      isReadOnly: ...
  }
}
```

app/components/rdf-input-fields/custom-field-wrapper.js
```diff
@@ -7,17 +7,26 @@ import { task } from 'ember-concurrency';
 import { validationsForFieldWithType } from '@lblod/submission-form-helpers';
 
 import { JSON_API_TYPE } from 'frontend-lmb/utils/constants';
+import { ADRES_CUSTOM_DISPLAY_TYPE } from 'frontend-lmb/utils/well-known-uris';
 
 export default class RdfInputFieldsCustomFieldWrapperComponent extends Component {
   @consume('on-form-update') onFormUpdate;
```
Rahien, line 13: see here we're now consuming three things as well! feels like it belongs together

Rahien (changes requested):

Rahien (approved):

Rahien (approved):

PR #464, opened 2025-01-20
lmb-1302 | Date input will also show a warning on end date

JonasVanHoof:
## Description

Change the severity of the min and max date to Warning instead of violation 

## How to test

- Go to a bestuursperiode 2019 -2024
- correct a mandataris 
- a warning should pop up when changing it to a date outside the period
- when wijzig mandataris is opened the save button should not be disabled because of the date

image

image

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/357


bfidlers (approved):
Works as expected however I noticed we allow entering end dates that are before the start date, might be something we want to look into, will post it in channel.

PR #467, opened 2025-01-22
lmb-1314: remove the disabled options for the bestuursperiode selector

JonasVanHoof:
## Description

When going to mandatarissen module the current bestuursperiode will be selected. When opening this may not be disabled when IV is active

## How to test

- All options show
- Disabled pill is gone

## Attachments

image

### DIFF
--- a/app/routes/mandatarissen/search.js
+++ b/app/routes/mandatarissen/search.js
@@ -26,7 +26,7 @@ export default class MandatarissenSearchRoute extends Route {
   };
 
   async model(params) {
-    const allBestuursperiode = await this.store.query('bestuursperiode', {
+    const allBestuursperioden = await this.store.query('bestuursperiode', {
       sort: 'label',
       include: [
         'installatievergaderingen',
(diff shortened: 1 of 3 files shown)


app/routes/mandatarissen/search.js
```diff
@@ -73,8 +56,8 @@ export default class MandatarissenSearchRoute extends Route {
 
     return {
       personenWithMandatarissen,
-      bestuursPeriods: periodMap,
-      selectedPeriod: { period: selectedPeriod, disabled: false },
+      allBestuursperiode,
```
Rahien, line 59: i think this should be either alleBestuursperioden or allBestuursperiodes

Rahien (approved):
just a small rename

app/routes/mandatarissen/search.js
JonasVanHoof: Changed to "allBestuursperioden"

JonasVanHoof: Used "allBestuursperiodes"

PR #471, opened 2025-01-27
clarify count warning for schepenen

Rahien:
## Description

image

## How to test

Change the start date of a schepen locally to somewhere in 2026


app/components/min-max-mandataris-warning.js
```diff
@@ -87,21 +88,31 @@ export default class MinMaxMandatarisWarningComponent extends Component {
         const min = mandaat.minAantalHouders;
         const max = mandaat.maxAantalHouders;
 
-        const label = (await mandaat.bestuursfunctie).label;
+        const bestuursfunctie = await mandaat.bestuursfunctie;
+        const label = bestuursfunctie.label;
+        const isSchepen = bestuursfunctie.uri === MANDAAT_SCHEPEN_CODE;
```
bfidlers, line 93: Is this only the schepenen mandaat? What about toegevoegde schepenen? If it is more than schepenen alone, I would use the getter from the mandaat model.

bfidlers (commented):
I'm a bit confused about it. The message says something about the preperation of the legislature, but this warning isn't in the prepare legislature page at all? Also it is not the case that we change the number of expected schepenen based on this right?
One more comment about the schepenen mandaat in the code.

If all this is as expected, I can approve since it does work...

app/components/min-max-mandataris-warning.js
Rahien: only for the schepen mandate, not for toegevoegde schepen

bfidlers (approved):
Works as expected. 
We need to think about how we hide this message whenever we are asked this, but those are issues for later.

Rahien:
> Works as expected. We need to think about how we hide this message whenever we are asked this, but those are issues for later.

we'll update the required counts

PR #474, opened 2025-01-29
lmb-1322 | update endpoint call to fetch nbMembers

JonasVanHoof:
## Description

We are now showing the amount of mandatarissen in the orgaan at "Aantal leden" this should be the amount of persons with an active mandaat.

## How to test

1. use the updated mandataris service
2. Go to organen
3. End a mandataris
4. The amount should have droppped by 1

Note: When you add a mandataris to an orgaan without mandatarissen this should be the easiest to see that it counts people.

## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/91

### DIFF
--- a/app/models/bestuursorgaan.js
+++ b/app/models/bestuursorgaan.js
@@ -182,18 +182,16 @@ export default class BestuursorgaanModel extends Model {
       return null;
     }
 
-    return Promise.all([
-      fetch(
-        `${API.MANDATARIS_SERVICE}/organen/${currentOrgaan.id}/activeMembers`
-      )
-        .then(async (response) => {
-          const result = await response.json();
-          return result?.count;
-        })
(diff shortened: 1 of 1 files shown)


app/models/bestuursorgaan.js
```diff
@@ -176,18 +178,22 @@ export default class BestuursorgaanModel extends Model {
...
-        return result.count;
-      })
-    );
-    const amount = mandatenAmounts.reduce((acc, curr) => acc + curr, 0);
-    return amount;
+    if (!currentOrgaan?.id) {
+      return null;
+    }
+
+    return Promise.all([
```
bfidlers, line 185: Why do you return a `Promise.all` here? Can't you just await this since this is an async function and you just do one call?

Rahien (approved):

PR #475, opened 2025-01-29
lmb-1338: becoming onafhankelijk is only for the current bestuursperi…

JonasVanHoof:
## Description

only become onafhankelijk for mandates in  current bestuursperiode 

## How to test

Become onafhankelijk ;p

image

### DIFF
--- a/app/routes/mandatarissen/persoon/mandaten.js
+++ b/app/routes/mandatarissen/persoon/mandaten.js
@@ -52,7 +52,6 @@ export default class MandatarissenPersoonMandatenRoute extends Route {
     return {
       persoon,
       hasActiveMandatarissen,
-      foldedMandatarissen,
       mandatarissenForOnafhankelijk,
       mandatarissen: filteredMandatarissen,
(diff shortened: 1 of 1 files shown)


app/routes/mandatarissen/persoon/mandaten.js
```diff
@@ -43,17 +44,33 @@ export default class MandatarissenPersoonMandatenRoute extends Route {
...
 
+    const mandatarissenForOnafhankelijk = await foldMandatarisses(
+      params,
+      await this.getMandatarissen(persoon, params, true)
+    );
+
     return {
       persoon,
       hasActiveMandatarissen,
       foldedMandatarissen,
```
bfidlers, line 55: would not pass these foldedMandatarissen anymore in the route, the are not used in template or controller, and this just makes it more confusing.

bfidlers (approved):
Small comment for cleaning it up a bit more. 
Otherwise looks good and seems to work!

PR #476, opened 2025-01-30
Ben/lmb 1339 fix fractie validation

bfidlers:
## Description

The mandataris-fractie-selector component failed in the case where there were multiple bestuursorganen selected e.g. for burgemeester mandates (both burgemeester and CBS bestuursorganen are selected in this case). 
This failing only happened when no fractie was filled in. 

This PR adds some extra safety checks that should handle this.

## How to test

Go to any mandataris which has a mandate with multiple bestuursorganen but has no fractie. Open the edit mistakes form, this should no longer throw an error.


JonasVanHoof (approved):
Looks good 👌🏻

PR #477, opened 2025-01-31
fixes after ember version bump

Rahien:
## Description

- fix using promise as ember record in update-state, need to await first 
- fix using promise as ember record in create-replacement, need to await first 
- fix missing inflector rule for library-entry
## How to test

- update state of mandataris, e.g. changing fraction
- update state of mandataris, select a replacement
- add custom field telephone number to form


bfidlers (approved):
I'm really confused why we have to await tracked variables? 
But it seems to fix the issues indeed.
Feels a bit slower, not sure if it is actually so or if it is just my machine doing other stuff.

PR #478, opened 2025-01-31
Lmb 1228 | v2 configurable instance columns

JonasVanHoof:
## Description

In the forms route we want to let the user configure what columns to view on an instance page.

## How to test

- Test together with both PR's (use yalc for semantic forms addon)
- Add some custom fields with a value to a schepen mandataris
- Go to the forms route and select the mandataris extra info module
- Play around with the columns

## Links to other PR's

- https://github.com/lblod/form-content-service/pull/71
- https://github.com/lblod/ember-semantic-forms/pull/4


app/templates/forms/form/instances.hbs
```diff
@@ -1,13 +1,36 @@
...
\ No newline at end of file
+
+<AuMainContainer as |main|>
+  <main.sidebar>
+    <div class="au-c-sidebar">
+      <div class="au-c-sidebar__header au-u-margin-top-small">
+        <AuHeading @level="3" @skin="3">Filters</AuHeading>
+      </div>
+      <div class="au-c-sidebar__content">
+        <InstanceTableConfiguration
```
Rahien, line 10: should this component not just be inside the InstanceTable on the side of the semantic-forms addon? That way this component will also be much more clean I think. You can just make it configurable with a parameter like `@configurableLabels` or something?

Rahien (changes requested):

Rahien (approved):

PR #479, opened 2025-02-03
fix mandataris list broken on some screen sizes around 1400

Rahien:
## Description

fix mandataris list broken on some screen sizes around 1400
## How to test

play around with screen size on orgaan/mandatarissen page and others
before:
image
after
image


JonasVanHoof (approved):
No more full sidebar around 1400

PR #482, opened 2025-02-04
don't always require fractie

Rahien:
## Description

don't always require fractie
## How to test

go to e.g. bcsd and add mandataris without fractie, go to gemeenteraad and see that you still can't

### DIFF
--- a/app/utils/form-context/application-context-meta-ttl.js
+++ b/app/utils/form-context/application-context-meta-ttl.js
@@ -94,5 +94,5 @@ export const loadIsFractieRequiredFromContext = (storeOptions) => {
     null,
     metaGraph
   );
-  return Literal.toJS(isRequired.value === '1' || isRequired.value === 'true');
+  return isRequired.value === '1' || isRequired.value === 'true';
 };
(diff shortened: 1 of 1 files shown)


bfidlers (approved):
I really don't get it, why before the verplicht label seemed correct, but the validation indeed failed. 

Now it works though.

bfidlers:
> I really don't get it, why before the verplicht label seemed correct, but the validation indeed failed.
> 
> Now it works though.

Okay investigated a bit further, the required label and the validation use different methods, which is a bit sketchy ...

app/utils/form-context/application-context-meta-ttl.js
```diff
@@ -94,5 +94,5 @@ export const loadIsFractieRequiredFromContext = (storeOptions) => {
     null,
     metaGraph
   );
-  return Literal.toJS(!!isRequired);
+  return Literal.toJS(isRequired.value === '1' || isRequired.value === 'true');
```
bfidlers, line 97: Does it still make sense to use the `Literal.toJS` here, because we do the check ourselves? I think we should either check if undefined, then return false otherwise return the `Literal.toJS` or just `isRequired.value === '1' || isRequired.value === 'true'` without the `Literal.toJS` around it.

Rahien: indeed!

bfidlers (approved):
Looks good!

PR #485, opened 2025-02-10
Ben/lmb 1319 improve burgemeester bekrachtiging

bfidlers:
## Description

Add warning when adding a burgemeester (Not really sure about the text and placement).
Screenshot 2025-02-10 at 09 51 07

Don't allow updating the publication status of burgemeester to bekrachtigd. This should be the case in the person page and in the bulk bekrachtiging. Also improved tooltips and the different cases in when the publication status can be edited.

## How to test

Try adding a burgemeester, this should still be possible, but should give you a warning.

You shouldnt' be able to set the publication status of a burgemeester to bekrachtigd (this should still be possible for aangewezen burgemeester). Test this in the bulk-bekrachtiging route and the persoon route and check if the messages are clear. Also test if other cases still work (RMW has no effectief state ...)

### DIFF
--- a/app/components/mandatarissen/mandataris-card.js
+++ b/app/components/mandatarissen/mandataris-card.js
@@ -78,7 +78,11 @@ export default class MandatarisCardComponent extends Component {
   }
 
   get lastStatusTooltipText() {
-    if (this.isEffectieLastStatus) {
+    if (this.args.legislatuurInBehandeling) {
+      return 'De publicatiestatus kan niet aangepast worden als de legislatuur nog in behandeling is.';
+    } else if (this.isEffectiefBurgemeester) {
+      return 'Een burgemeester kan enkel bekrachtigd worden via een benoeming.';
+    } else if (this.isEffectiefLastStatus) {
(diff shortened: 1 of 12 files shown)


app/components/mandatarissen/mandataris-card.js
```diff
@@ -51,18 +77,12 @@ export default class MandatarisCardComponent extends Component {
...
-      this.args.mandataris.publicationStatus?.get('id');
-    if (
-      this.effectiefIsLastStatus &&
-      publicatieStatusId &&
-      publicatieStatusId === PUBLICATION_STATUS_EFFECTIEF_ID
-    ) {
+    if (this.args.legislatuurInBehandeling) {
+      return 'De publicatiestatus kan niet aangepast worden als de legislatuur nog in behandeling is.';
+    } else if (this.isEffectiefBurgemeester) {
+      return 'Een burgemeester kan enkel bekrachtigd worden via een benoeming.';
```
Rahien, line 84: (dit gebeurt automatisch)

Rahien (changes requested):

PR #488, opened 2025-02-11
fix manueel bekrachtigd links

Rahien:
## Description

Allow users to fix manually bekrachtigings links

image

image

image

## How to test

try it, gr gent is a good place


bfidlers (approved):
Works as expected!

Small remark, I would only allow saving when you have actual changes, this would be more consistent with the rest of our application. But it does not make an actual differency in functionality.

PR #493, opened 2025-02-14
Reroute when trying to access disabled legislature page

bfidlers:
## Description

Also disable the route when the legislatuur is hidden.

## How to test

Manually try to go to [prepare legislature page](http://localhost:4200/verkiezingen/installatievergadering) for a bestuurseenheid where it is disabled. See if you got rerouted to the index page.

### DIFF
--- a/app/routes/verkiezingen.js
+++ b/app/routes/verkiezingen.js
@@ -7,11 +7,16 @@ import { BESTUURSEENHEID_CLASSIFICATIECODE_GEMEENTE } from 'frontend-lmb/utils/w
 export default class VerkiezingenRoute extends Route {
   @service currentSession;
   @service session;
+  @service router;
 
   beforeModel(transition) {
     this.session.requireAuthentication(transition, 'login');
 
     if (!this.currentSession.canAccessMandaat) {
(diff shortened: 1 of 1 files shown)


app/routes/verkiezingen.js
```diff
@@ -7,13 +7,18 @@ import { BESTUURSEENHEID_CLASSIFICATIECODE_GEMEENTE } from 'frontend-lmb/utils/w
...
   beforeModel(transition) {
     this.session.requireAuthentication(transition, 'login');
 
     if (!this.currentSession.canAccessMandaat) {
       this.router.transitionTo('index');
     }
+
+    if (!this.currentSession.showLegislatuurModule) {
+      this.router.transitionTo('index');
+    }
```
JonasVanHoof, line 21: Nitpick: would do it like this, the outcome of the if statement is the same, but I do like the readability of your version (if it stays for two check ;p)  

```suggestion
    if (
      !this.currentSession.canAccessMandaat ||
      !this.currentSession.showLegislatuurModule
    ) {
      this.router.transitionTo('index');
    }
```

JonasVanHoof (approved):

PR #494, opened 2025-02-14
Ben/lmb 1245 locally managed

bfidlers:
## Description

Disable application pages when a bestuurseenheid is locally managed.

## How to test

First checkout out the backend. And restart resources. Then you need to add a predicate to a bestuurseenheid to mark it as locally managed. The following query does this for Gemeente Brugge, but you can choose your own bestuurseenheid.

```
  PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
  PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>
  PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
    INSERT {
      GRAPH ?h {
        ?bestuurseenheid ext:lokaalBeheerd "true"^^xsd:boolean .
      }
    }
    WHERE {
      VALUES ?bestuurseenheid { <http://data.lblod.info/id/bestuurseenheden/28346950e285b8b816133fece5ac9408097c3f190c7f32573cf0c640d6c34b1a> }
      GRAPH ?g {
        ?bestuurseenheid a besluit:Bestuurseenheid .
      }
      BIND(?g AS ?h)
    }

```
After you ran this query you can go to the specific bestuurseenheid. And from the moment you login you should be redirected to an info page about the locally managed stuff. You shouldn't be able to see any other pages that are directly available (e.g. links you can click on or the major modules).

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/383

### DIFF
--- a/app/services/current-session.js
+++ b/app/services/current-session.js
@@ -124,7 +124,6 @@ export default class CurrentSessionService extends Service {
   }
 
   get isLokaalBeheerd() {
-    console.log(this.group);
     return this.group.lokaalBeheerd;
   }
 
--- a/app/services/current-session.js
+++ b/app/services/current-session.js
@@ -123,6 +123,11 @@ export default class CurrentSessionService extends Service {
     return this.canAccess(MODULE.MANDATENBEHEER);
   }
 
+  get isLokaalBeheerd() {
+    console.log(this.group);
+    return this.group.lokaalBeheerd;
+  }
+
   get isAdmin() {
     let roles = this.roles;
     if (this.impersonation.isImpersonating) {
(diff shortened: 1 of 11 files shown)


JonasVanHoof (approved):
Je kan geen kant meer op!

image

app/services/current-session.js
```diff
@@ -123,6 +123,11 @@ export default class CurrentSessionService extends Service {
     return this.canAccess(MODULE.MANDATENBEHEER);
   }
 
+  get isLokaalBeheerd() {
+    console.log(this.group);
```
JonasVanHoof, line 127: remove this log

PR #507, opened 2025-02-27
LMB 1353 | Make rangorde a dropdown

JonasVanHoof:
## Description

Instead of a text input field we want the rangorde to be shown as a dropdown with the options 1-...100 so the user can only select a valid rangorde. The up and down arrows stay visible next to the dropdown so they can go up and down with automatic reordering of the mandataris with rangorde.

## How to test

Do these test in edit-rangorde (organen) and in the IV page

1. clear a rangorde with the dropdown
2. assign a rangorde through the dropdown
3.  use the up/down arrows to change the rangorde value (see that the previous rangorde value is being replaced)
4. can set a duplicate rangorde by using the dropdown (not the up/down arrows)

## Attachements

**On IV page**
image

**In the mandataris form**
image

**On the edit-rangorde page**
image

### DIFF
--- a/app/routes/organen/orgaan/edit-rangorde.js
+++ b/app/routes/organen/orgaan/edit-rangorde.js
@@ -6,6 +6,7 @@ import { endOfDay } from 'frontend-lmb/utils/date-manipulation';
 
 export default class EditRangordeRoute extends Route {
   @service store;
+  @service bestuursperioden;
   @service('mandatarissen') mandatarissenService;
 
   queryParams = {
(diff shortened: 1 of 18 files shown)


app/routes/organen/orgaan/edit-rangorde.js
```diff
@@ -37,11 +38,17 @@ export default class EditRangordeRoute extends Route {
       return { mandataris: mandataris, rangorde: mandataris.rangorde };
     });
 
+    const currentBestuursperiode =
+      await this.bestuursperioden.getCurrentBestuursperiode();
+    const isEditingInPast =
+      parentModel.selectedBestuursperiode.id !== currentBestuursperiode.id;
+
```
JonasVanHoof, line 45: Added this as wijzig mandatarissen for a period other than the current period it shows multiple mandatarissen in the table

image

JonasVanHoof: image

Rahien: haven't checked the code/reasoning for disabling this yet but the tooltip shown doesn't make sense. 2019-2024 is in the past. And i believe we should allow correcting rangorde both past, present and future

JonasVanHoof: Reverted this commit

JonasVanHoof: Fixed by Karel

PR #512, opened 2025-03-06
new end dates

Rahien:
## Description

End of day is now same day but 23h59:59
## How to test

update a mandataris (state and correct) and check the end date, it should be displayed correctly in the interface and it should be set to the same day at 23h59:59

### DIFF
--- a/app/utils/date-manipulation.js
+++ b/app/utils/date-manipulation.js
@@ -2,12 +2,18 @@ import moment from 'moment';
 
 export function endOfDay(date) {
   if (date) {
-    return moment(date).add(1, 'days').startOf('day').utc().toDate();
+    return moment(date)
+      .add(1, 'days')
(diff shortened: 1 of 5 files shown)


app/utils/date-manipulation.js
```diff
@@ -2,12 +2,18 @@ import moment from 'moment';
 
 export function endOfDay(date) {
   if (date) {
-    return moment(date).add(1, 'days').startOf('day').utc().toDate();
+    return moment(date)
+      .add(1, 'days')
+      .startOf('day')
+      .subtract(1, 'seconds')
```
JonasVanHoof, line 8: Nice and simple fix!

JonasVanHoof (approved):
When an end date is set it displays the same day at 23:59:59

image

PR #513, opened 2025-03-07
DO NOT MERGE: check before creating person

Rahien:
## Description

PR to show the path where we check before creating a person through a form. This feels wrong and as Ben said, it's probably a case where we should ask for forgiveness rather than for permission.


Rahien:
closed in favour of doing this as a cron and cleaning up afterward: https://github.com/lblod/merge-person-service/pull/3

PR #520, opened 2025-03-24
basic timeline for history

Rahien:
## Description

Show the basic timeline for the history of a mandataris, allow switching events by clicking the cards or the dates (allowing for keyboard only interaction too)

image

Scroll on smaller screens:
image

only basics are shown for now and the feature is behind a feature flag. Advanced things like what changed and corrections are separate stories.

## How to test

Go to gent and open a mandataris with a high rank, play around with correcting and updating state.

### DIFF
--- a/app/components/mandatarissen/timeline.js
+++ b/app/components/mandatarissen/timeline.js
@@ -7,12 +7,11 @@ export default class MandatarissenTimeline extends Component {
   @service store;
 
   get isFeatureEnabled() {
-    return this.features.isEnabled('shacl-report');
+    return this.features.isEnabled('timeline');
   }
 
   get timelineEvents() {
-    const mandatarissen = this.args.mandatarissen;
-    const sortedMandatarissen = mandatarissen.sortBy('start');
+    const sortedMandatarissen = this.args.mandatarissen.sortBy('start');
     const events = sortedMandatarissen.map((mandataris, index) => {
       return {
         type: index > 0 ? 'Wijziging' : 'Start',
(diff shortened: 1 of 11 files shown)


app/components/mandatarissen/timeline.js
```diff
@@ -0,0 +1,45 @@
+import Component from '@glimmer/component';
+import { service } from '@ember/service';
+import moment from 'moment';
+
+export default class MandatarissenTimeline extends Component {
+  @service features;
+  @service store;
+
+  get isFeatureEnabled() {
+    return this.features.isEnabled('shacl-report');
```
JonasVanHoof, line 10: Is this the correct feature flag?
```suggestion
    return this.features.isEnabled('timeline');
```

app/components/mandatarissen/timeline.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  get timelineEvents() {
+    const mandatarissen = this.args.mandatarissen;
+    const sortedMandatarissen = mandatarissen.sortBy('start');
```
JonasVanHoof, line 15: The unsorted mandatarissen are never used after sort was applied on  it
```suggestion
const sortedMandatarissen = this.args.mandatarissen.sortBy('start');
```

JonasVanHoof (changes requested):
I like it! 

* Just change the feature flag name
* For me the timeline and the old historiek shows twice, tried playing with the feature flag but it only seems to just toggle the visibility of the new timeline. Not sure how thats possible has you have put it in an `if else` ;p
image

Rahien:
> For me the timeline and the old historiek shows twice, tried playing with the feature flag but it only seems to just toggle the visibility of the new timeline. Not sure how thats possible has you have put it in an if else ;p

As this is not a full release of the historiek feature yet but more a work in progress, i wanted to keep the old version there. It's easy to work with and compare the results. (and we'll need it when adding the diffs and warnings to easily see whether we still have the same thing). It can be removed once the feature is ready to ship

JonasVanHoof (approved):
Great!

PR #523, opened 2025-03-28
Ben/lmb 1310 end date validation

bfidlers:
## Description

Add validation to mandataris start and end date, which prevents the end date being before the start date.

I needed to add two validations, one to the start field and one to the end field. First I only added a greaterThan validation to the mandaat:einde, but in that case if you first fill in the end date and only after that the start date, you never saw the validation, but the form was not valid, this was pretty confusing. 

This is the warning shown when you first fill in the start date and then the end date:
Screenshot 2025-03-28 at 01 48 14

This is the warning shown when you do it the other way around:
Screenshot 2025-03-28 at 01 49 04

There is one edge case which is not ideal, if you have a warning on the end field but then change the start field to a valid value, the warning does not disappear (you can save the form however). There are a few ways to solve this, but I don't think they are worth it, they would be pretty substantial to how the forms work or need dedicated components for the start and end field, which seems overkill.
Screenshot 2025-03-28 at 01 48 38

## How to test

Create a new mandataris and try adding an end date which is before the start date, this should not be possible.

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/409


app/utils/form-validations/greater-than.js
```diff
@@ -0,0 +1,30 @@
...
+
+  const threshold = getValue(comparisonPredicate, options);
+
+  if (!inputValue) {
+    return true;
+  }
+  if (!threshold) {
+    return false;
+  }
+  return inputValue.value >= threshold.value;
```
JonasVanHoof, line 19: Use Literal.toJS to compare

JonasVanHoof: ref on how this could work https://github.com/lblod/ember-submission-form-fields/blob/a21c0df3c657a7e99fb95999e7cee3d23cd6dbcf/addon/models/listing.js#L103

JonasVanHoof (dismissed):

app/utils/form-validations/greater-than.js
```diff
@@ -0,0 +1,30 @@
+import { EXT } from 'frontend-lmb/rdf/namespaces';
+
+export function greaterThan([inputValue], options) {
```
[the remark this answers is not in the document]
bfidlers: Added a datetype to the validation, that way we can fairly simply distingish between different types of literals.

Rahien (approved):

PR #524, opened 2025-03-28
LMB-1218 | Handle removing custom forms

JonasVanHoof:
## Description

We want to remove the custom forms and there instances. 

## How to test

Delete a form or instance and see if you where given enough information to confidently remove the form.

## Links to other PR's

- https://github.com/lblod/form-content-service/pull/76

### DIFF
--- a/app/controllers/custom-forms/instances/index.js
+++ b/app/controllers/custom-forms/instances/index.js
@@ -101,14 +101,13 @@ export default class CustomFormsInstancesIndexController extends Controller {
 
   get removeFormModalText() {
     if (this.model.usage.hasUsage) {
-      const usageCount = this.model.usage.usageUris.length;
       let text = (variableText) =>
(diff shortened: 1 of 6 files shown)


app/controllers/custom-forms/instances/index.js
```diff
@@ -59,6 +67,50 @@ export default class CustomFormsInstancesIndexController extends Controller {
...
+      );
+    }
+    this.isDeleting = false;
+    this.isDeleteModalOpen = false;
+    this.formToDelete = null;
+  }
+
+  get removeFormModalText() {
+    if (this.model.usage.hasUsage) {
+      const usageCount = this.model.usage.usageUris.length;
```
Rahien, line 104: would just return the count, else this list can be very long as it's not paginated

Rahien (changes requested):

PR #525, opened 2025-03-28
Karel/lmb 1517 how to show form links

Rahien:
## Description

This allows users to specify which fields to show in summary views (instance tables, links)
Note: this only allows configuring these fields and using them in instance tables. It also does not take into account fancy types like addresses or nested values. That feels like its own separate thing

## How to test
use yalc to have the linked semantic-forms version
run the linked form-content service instead of the regular one in the app

Create a custom form
Add a field
Add another field, mark it as show in summary
Add another field
Make the first field also show in summary
Add some instances to the form
See that by default the fields are shown in the list 

Make the second field no longer show in the summary
See that by default the field is no longer shown in the list of instances

See that you can still configure the columns in the list of instances
See that you can't make a field show up in a summary for form extensions
## Links to other PR's

- https://github.com/lblod/ember-semantic-forms/pull/6
- https://github.com/lblod/form-content-service/pull/75
- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/412

### DIFF
--- a/app/components/rdf-input-fields/crud-custom-field-modal.js
+++ b/app/components/rdf-input-fields/crud-custom-field-modal.js
@@ -15,6 +15,7 @@ import {
   LIBRARY_ENTREES,
   TEXT_CUSTOM_DISPLAY_TYPE,
 } from 'frontend-lmb/utils/well-known-uris';
+import { Literal, NamedNode } from 'rdflib';
 
 export default class RdfInputFieldCrudCustomFieldModalComponent extends Component {
   @consume('form-context') formContext;
@@ -224,7 +225,17 @@ export default class RdfInputFieldCrudCustomFieldModalComponent extends Componen
 
   get isCustomForm() {
     const forkingStore = this.forkingStore;
-    return !forkingStore.any(null, EXT('extendsForm'), null, SOURCE_GRAPH);
+    return (
+      forkingStore.any(
+        null,
+        EXT('isCustomForm'),
+        new Literal(
+          'true',
+          null,
+          new NamedNode('http://www.w3.org/2001/XMLSchema#boolean')
+        )
+      ) && !forkingStore.any(null, EXT('extendsForm'), null, SOURCE_GRAPH)
+    );
   }
 
   get libraryFieldOptions() {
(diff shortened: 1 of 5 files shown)


JonasVanHoof (changes requested):
When I add a form field i see the summary toggle but when adding another one the form isn/t custom anymore ?

app/components/rdf-input-fields/crud-custom-field-modal.js
```diff
@@ -203,14 +213,33 @@ export default class RdfInputFieldCrudCustomFieldModalComponent extends Componen
...
+    return (
+      forkingStore.any(
+        null,
+        EXT('isCustomForm'),
+        new Literal(
+          'true',
+          null,
+          new NamedNode('http://www.w3.org/2001/XMLSchema#boolean')
+        )
+      ) && !forkingStore.any(null, EXT('extendsForm'), null, SOURCE_GRAPH)
```
JonasVanHoof, line 237: Ah this one,indeed!

app/components/rdf-input-fields/crud-custom-field-modal.js
```diff
@@ -203,14 +213,33 @@ export default class RdfInputFieldCrudCustomFieldModalComponent extends Componen
...
       SOURCE_GRAPH,
       'text/turtle'
     );
+    return forkingStore;
+  }
+
+  get isCustomForm() {
+    const forkingStore = this.forkingStore;
+    return (
+      forkingStore.any(
```
JonasVanHoof, line 229: Would just use `this.forkingStore` instead of making a const out of it?

app/components/shared/compact-menu.hbs
```diff
@@ -20,6 +20,12 @@
         Opstart nieuwe legislatuur
       </AuLink>
     {{/if}}
+    {{#if (is-feature-enabled "show-eigen-gegevens-module")}}
```
JonasVanHoof, line 23: These changed* => will be fixed with the rebase

app/utils/constants.js
```diff
@@ -21,9 +21,11 @@ export const RESOURCE_CACHE_TIMEOUT = 1000;
 export const API = {
   MANDATARIS_SERVICE: '/mandataris-api',
   CONCEPT_SCHEME_SERVICE: '/concept-scheme-api',
+  FORM_CONTENT_SERVICE: '/form-content',
 };
 export const STATUS_CODE = {
   OK: 200,
+  CREATED: 201,
```
JonasVanHoof, line 28: Not sure how Ben did the error-handling but maybe we should just check `response.ok` instead of a specific status?

JonasVanHoof (changes requested):
Looks good!
image

Note: when changing the field in summary in the form defintion modal it needs a manual refresh to show => very acceptable to me


Adding request changes for the rebase

JonasVanHoof (commented):
Looks good 👍🏻

JonasVanHoof (approved):

PR #526, opened 2025-03-28
make it clear when the mandataris shown is not the latest version

Rahien:
## Description

make it clear when the mandataris shown is not the latest version
image

## How to test

go to a mandataris that has changes, pick an old state, see the badge and the tooltip


JonasVanHoof (approved):
clean 🚀

PR #529, opened 2025-04-01
remove bad tests

Rahien:
## Description

removes auto-generated tests that fail (e.g. because they expect an argument or some state in a service).
This way we can start from a more or less clean slate of working tests and add tests that we feel are important (not some auto-generated stuff that breaks all the time)
## How to test

 ember test --launch firefox


tests/unit/utils/rijksregisternummer-test.js
```diff
@@ -39,14 +39,3 @@ module(
...
-module(
-  'Unit | Utils | form-validations | RRN helpers | 06.02.02-000.80',
-  function () {
-    const rrn = '06020200080';
-    const formattedRrn = '06.02.02-000.80';
-    test(`${formattedRrn} is an invalid RRN, serial should be between 001 and 998`, function (assert) {
-      assert.false(isValidRijksregisternummer(rrn));
-    });
-  }
-);
```
JonasVanHoof, line 52: Mijn mooie test..

JonasVanHoof (approved):
All test pass, nice cleanup! For me personally the unit test folder can go aswell, now it looks like we do have unit test :p

PR #531, opened 2025-04-01
Ben/lmb 1476 rework edit mandataris

bfidlers:
## Description

Reworked edit mandataris form. This is in a new route, now there is only one form and the second page is now a selector where you decide if you want to update state or correct mistakes, in the future this second page will have more functionality.

## How to test

Check if the reworked page is behind a feature flag. `feature-edit-mandataris-rework=true` to show the reworked page.

Check if you can still edit mandatarissen and it the effects are similar as before. There are some small differences because everything is now in one form and in the future we will decide if the user wants to perform an update state or a corrigeer fouten.


app/components/mandataris/edit/form.js
```diff
@@ -0,0 +1,333 @@
...
+
+  @tracked correctedMandataris = false;
+  @tracked updatedStateMandataris = false;
+  @tracked newMandataris;
+
+  @tracked isRangordeModalOpen;
+
+  constructor() {
+    super(...arguments);
+    this.load.perform();
```
Rahien, line 53: can't we do this with a tracked function

app/components/mandataris/edit/mandaat-selector.js
```diff
@@ -0,0 +1,95 @@
...
+  @tracked errorValidations = [];
+  @tracked warningValidations = [];
+  @tracked initialized = false;
+  @tracked hasBeenFocused = false;
+  @tracked mandaat = null;
+  @tracked mandaatOptions = null;
+
+  constructor() {
+    super(...arguments);
+    this.load();
```
Rahien, line 20: can we use a tracked function instead?

Rahien (changes requested):

app/templates/mandatarissen/persoon/mandataris-rework.hbs
```diff
@@ -0,0 +1,116 @@
...
+            <span>Beheerd door:</span>
+            <Mandaat::OwnershipTags @owners={{this.model.owners}} />
+          </div>
+          <p>Je kan deze mandataris niet bewerken.</p>
+        </AuAlert>
+      {{else}}
+        <AuButtonGroup>
+          <Shared::Tooltip
+            @showTooltip={{true}}
+            @tooltipText={{this.toolTipText}}
```
Rahien, line 31: shouldn't show the tooltip about the prep of the next period if that prep is already done

app/components/mandataris/edit/form.js
```diff
@@ -0,0 +1,333 @@
...
+      return false;
+    }
+    if (this.replacement) {
+      this.isRangordeModalOpen = true;
+    }
+  }
+
+  @action
+  callbackAfterUpdate() {
+    if (this.newMandataris != this.model.mandataris) {
```
Rahien, line 326: this.model is undefined

PR #533, opened 2025-04-02
LMB-1332 | Extend prototypes false

JonasVanHoof:
## Description

We need to set this in the environment.js file. With this comes some updates of array methods.
```js
    EmberENV: {
      EXTEND_PROTOTYPES: false,
```

## How to test

Everything should work as before. Definitely focus on reactivity!

### DIFF
--- a/app/components/address/adressenregister-selector.js
+++ b/app/components/address/adressenregister-selector.js
@@ -36,7 +36,9 @@ export default class AdressenregisterSelectorComponent extends Component {
         const selectedAddress = addresses.find(
           (a) => a.busnumber == address.busnummer
         );
-        this.addressesWithBusnumbers = addresses.sortBy('busnumber');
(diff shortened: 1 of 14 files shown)


app/components/address/adressenregister-selector.js
```diff
@@ -36,7 +36,9 @@ export default class AdressenregisterSelectorComponent extends Component {
         const selectedAddress = addresses.find(
           (a) => a.busnumber == address.busnummer
         );
-        this.addressesWithBusnumbers = addresses.sortBy('busnumber');
+        this.addressesWithBusnumbers = [...addresses].sort(
```
JonasVanHoof, line 39: SO ugly :/

Rahien: yeah sort never should have been destructive but here we are

Rahien (approved):

PR #535, opened 2025-04-07
show corrections in timeline

Rahien:
## Description

- show correction number and fix corrections not linked to right mandataris
- show collapsible corrections (unless only 1 being the creation of the instance)

image
image

## How to test

make some corrections and state changes to a mandataris, see that the corrections all show up and that the time and owner are shown for all of the mandatarisses in the timeline

### DIFF
--- a/app/components/mandatarissen/timeline-corrections.js
+++ b/app/components/mandatarissen/timeline-corrections.js
@@ -20,7 +20,7 @@ export default class MandatarissenTimelineCorrections extends Component {
   }
 
   get showCorrections() {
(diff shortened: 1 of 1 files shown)


bfidlers:
Didn't look at the code yet, but I had some weird states. 

1. It only showed corrections when there were two corrections and the creation. If I only corrected ones, it did not show.
Screenshot 2025-04-07 at 12 33 12
2. Is it correct that it only shows when it changes happened and not what the changes were?
3. Got this weird state, but not sure how it happened.
Screenshot 2025-04-07 at 12 35 29

bfidlers (approved):
Looks good

PR #536, opened 2025-04-09
LMB-1502 | Allow linking to other form instances

JonasVanHoof:
## Description

We want a new custom field where the user can add a link to another form (default or custom forms). 

## How to test

1. Go to eigen gegeven and add a new form
2. Add a field "Link to formulier" to the form definition
3. See if all options are shown in the dropdown as expected (show in summary predicate)
5. Create a link between this form and the one you selected
6. This  new field can also be shown in summary for the instance table => URI

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/416
- https://github.com/lblod/form-content-service/pull/78

### DIFF
--- a/app/components/custom-form/link-to-form-instance.js
+++ b/app/components/custom-form/link-to-form-instance.js
@@ -21,15 +21,16 @@ export default class CustomFormLinkToFormInstance extends SelectorComponent {
   @service semanticFormRepository;
 
   @use(getAllFormLabels) getAllFormLabels;
-  @use(getInstances) getInstances;
   @use(getInitialSelectedInstances) getInitialSelectedInstances;
 
   @consume('form-context') formContext;
 
+  @tracked instanceObjectsOfOptions = [];
   @tracked extraInstanceUris = [];
   @tracked selectedInstances = A([]);
   @tracked instanceDisplayLabels = [];
-  @tracked pageToLoad = 0;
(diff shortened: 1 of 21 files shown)


bfidlers (commented):
It works, but I don't really see the use case where you want to allow the user to add a link to any instance they want. If this is by demand, then it is fine and it works as expected.

Don't you want to configure the type of instance when you edit the form and when you create instances, then you can only select links to the instance type you selected?

Otherwise looks pretty good.

app/components/custom-form/link-to-form-instance.js
```diff
@@ -0,0 +1,210 @@
...
+        (m) =>
+          !selectedInstances.find((option) => m.value === option.instance.uri)
+      )
+      .forEach((m) => updateSimpleFormValue(this.storeOptions, undefined, m));
+    selectedInstances
+      .filter((option) => !matches.find((m) => option.instance.uri === m.value))
+      .forEach((option) =>
+        updateSimpleFormValue(
+          this.storeOptions,
+          new Literal(option.instance.uri)
```
Rahien, line 118: this should be namednode i think

JonasVanHoof: You are correct 🙏🏻

Rahien (approved):

PR #537, opened 2025-04-09
hide add custom organs

Rahien:
## Description

put ~custom organen~ adding new persons to custom organen behind feature flag.
## How to test

disable the feature flag (enabled by default in dev) and see that the you can't add new persons to organ if it is custom

### DIFF
--- a/app/services/bestuursorganen.js
+++ b/app/services/bestuursorganen.js
@@ -49,10 +49,8 @@ export default class BestuursorganenService extends Service {
     if (queryParams.activeOrgans) {
       queryOptions['filter[:has-no:deactivated-at]'] = true;
     }
-    let filteredTypes = ['decretaleIds'];
-    if (this.features.isEnabled('custom-organen')) {
-      filteredTypes = queryParams.selectedTypes;
-    }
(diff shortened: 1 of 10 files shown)


bfidlers (commented):
The feature flag works, and it is indeed no longer possible to add new custom organen or to show them in the general orgaan page. There are however still a few ways of getting there. You can still select the organen when you add a new mandate to a person, and the current mandates in the custom organen are still shown in the mandatarissen view. 
But it all depends how far do we want to go here...

app/services/bestuursorganen.js
```diff
@@ -6,6 +6,7 @@ export default class BestuursorganenService extends Service {
   @service store;
   @service decretaleOrganen;
   @service bestuursperioden;
+  @service features;
```
bfidlers, line 9: not used

bfidlers (approved):
It works but it feels a bit weird.

The check that is added to the search-form is `hasElections` but this actually has nothing to do with elections? And the feature is called `custum-organen` but it is not really about the organen themselves, but whether they allow creating people or not.

PR #543, opened 2025-04-17
sanding off validations

Rahien:
## Description

Better status when loading validations, fewer notifications if not necessary.
Put all validation errors in a single table, give the option of showing more details

image
image

## How to test

see linked pr
## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/419

### DIFF
--- a/app/components/refresh-validations-button.hbs
+++ b/app/components/refresh-validations-button.hbs
@@ -2,15 +2,7 @@
   <div class="report-refresh-button au-u-flex">
     <AuPill @icon="clock-rewind">Laatste sync
       {{this.lastRefreshedDate}}</AuPill>
-    <Shared::Tooltip
-      @showTooltip={{and
-        this.isPreviousreportFound
-        this.tooltipLastSyncText
-        (not this.isGeneratingReport)
-      }}
-      @alignment="right"
-      @tooltipText={{this.tooltipLastSyncText}}
-    >
+    <Shared::Tooltip @alignment="right">
       <AuButton
         @disabled={{this.isGeneratingReport}}
         @skin="naked"
--- a/app/components/refresh-validations-button.hbs
+++ b/app/components/refresh-validations-button.hbs
@@ -2,19 +2,17 @@
   <div class="report-refresh-button au-u-flex">
     <AuPill @icon="clock-rewind">Laatste sync
       {{this.lastRefreshedDate}}</AuPill>
-    <Shared::Tooltip @alignment="right">
-      <AuButton
-        @disabled={{this.isGeneratingReport}}
-        @skin="naked"
-        class="refresh-button-next-to-pill"
-        {{on "click" this.refresh}}
-      >
-        <AuIcon
-          @icon="renew"
-          class={{if this.isGeneratingReport "rotate-element-360-degree"}}
-        />
-        Refresh</AuButton>
-    </Shared::Tooltip>
+    <AuButton
+      @disabled={{this.isGeneratingReport}}
+      @skin="naked"
+      class="refresh-button-next-to-pill"
+      {{on "click" this.refresh}}
+    >
+      <AuIcon
+        @icon="renew"
+        class={{if this.isGeneratingReport "rotate-element-360-degree"}}
+      />
+      Refresh</AuButton>
   </div>
   <p class="au-u-flex au-u-flex-self-end au-u-padding-top-small">Meer weten over
     validatie fouten?</p>
--- a/app/components/validation-report-row.hbs
+++ b/app/components/validation-report-row.hbs
@@ -1,0 +1,23 @@
+<td>
+  <LinkTo
+    @route={{@result.context.route}}
+    @model={{@result.context.modelId}}
+    class="au-c-link au-u-medium"
+  >
+    {{@result.label}}
+  </LinkTo>
+</td>
+<td>
+  <span>{{@result.message}}</span>
+  {{#if this.collapsed}}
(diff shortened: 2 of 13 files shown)


app/components/refresh-validations-button.hbs
```diff
@@ -2,15 +2,7 @@
...
-    <Shared::Tooltip
-      @showTooltip={{and
-        this.isPreviousreportFound
-        this.tooltipLastSyncText
-        (not this.isGeneratingReport)
-      }}
-      @alignment="right"
-      @tooltipText={{this.tooltipLastSyncText}}
-    >
+    <Shared::Tooltip @alignment="right">
```
JonasVanHoof, line 5: No text for this tooltip? I would leave the tooltip here. This will make it possible for the user to just peek how long the last validation took when there already is a generated report 💭

app/components/validation-report-row.hbs
```diff
@@ -0,0 +1,23 @@
...
+<td>
+  <span>{{@result.message}}</span>
+  {{#if this.collapsed}}
+    <AuButton
+      @skin="link"
+      {{on "click" this.toggleCollapsed}}
+    >[detail]</AuButton>
+  {{else}}
+    <span class="au-u-muted">{{@result.detail}}</span>
+    <AuButton @skin="link" {{on "click" this.toggleCollapsed}}>[verberg detail]</AuButton>
```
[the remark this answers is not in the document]
JonasVanHoof: I do like the more details 😃  !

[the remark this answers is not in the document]
JonasVanHoof: just always show the details below so the show and hide button is not jumping around

app/templates/report.hbs
```diff
@@ -19,16 +19,24 @@
...
+              pagina herlaadt automatisch wanneer het rapport klaar is en u kan
+              ondertussen de applicatie blijven gebruiken.
+            </p>
+            {{#if this.validatie.lastValidationDuration}}
+              <p>De vorige validatie duurde
+                {{this.validatie.lastValidationDuration}}</p>
+            {{/if}}
+          </AuAlert>
+        {{else}}
+          <div class="au-u-box no-pagination">
```
JonasVanHoof, line 35: Add some paddin gor margin to the bottom so it keeps completly visible at all times
image

My screenshot is a little hardcore xd it wasn't this bad on my other screens*

app/models/bestuurseenheid.js
```diff
@@ -45,7 +45,7 @@ export default class Bestuurseenheid extends Model {
 
   @hasMany('bestuursorgaan', {
     async: true,
-    inverse: 'orginalBestuurseenheid',
+    inverse: 'originalBestuurseenheid',
```
JonasVanHoof, line 48: Damn had to take a good look to see that 😅  good job!

app/templates/report.hbs
```diff
@@ -19,16 +19,24 @@
...
           </AuAlert>
         {{/if}}
-        {{#each this.resultsByTargetClass as |instanceGroup|}}
-          <div class="au-u-box au-u-margin-top no-pagination">
-            <AuHeading @skin="3" class="au-u-margin-bottom">
-              {{instanceGroup.class.label}}
-            </AuHeading>
-
+        {{#if this.reportIsRunning}}
+          <AuAlert @skin="info" @icon="clock">
```
JonasVanHoof, line 23: Should we also add the start time so they can reflect if it is almost done?

app/components/validation-report-row.hbs
```diff
@@ -0,0 +1,23 @@
+<td>
```
JonasVanHoof, line 1: Probably I did it wrong but i got this validation report error
image

Clicked the linked mandataris and did `WIJZIG mandataris` ran validations again and it wasn't fixed as obiously I had to correct the mandataris 🤷🏻  probably also fixed with the mandataris wizard? Or should we only get the active mandatarissen in the sparqlShapes?

app/components/validation-report-row.hbs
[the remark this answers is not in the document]
JonasVanHoof: Hm I see why this could be difficult haha just show it below the message than?

app/templates/report.hbs
```diff
@@ -19,16 +19,24 @@
...
+            <p>
+              Het validatierapport wordt gegenereerd. Dit kan even duren. Deze
+              pagina herlaadt automatisch wanneer het rapport klaar is en u kan
```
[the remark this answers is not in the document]
JonasVanHoof: It did the second time, now it did 🤔  This is because it breaks with the currenstatus Id in the comment before

app/components/validation-report-row.hbs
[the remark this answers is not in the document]
Rahien: damn that is one shitty detail message, will improve!

app/templates/report.hbs
Rahien: reports now run almost instantly, i don't think there is any need

app/components/validation-report-row.hbs
Rahien: wizard should make it so you don't have to think about this and it will decide that you should have picked correct mistakes for you

app/components/refresh-validations-button.hbs
Rahien: i left it out as reports now run almost instantly and i found it to be a bit too distracting with the tooltip

PR #544, opened 2025-04-18
LMB-1460 | Rework edit mandataris validations

JonasVanHoof:
## Description

Now that the form for editing a mandataris is an "ember-way" form we need to rework the validations for this.

- The layout is consistent across the application and between different fields
- required is supported and used
- end date can’t be before start date 

## How to test

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

## Links to other PR's

- /


app/utils/date.js
```diff
@@ -0,0 +1,28 @@
+import moment from 'moment';
+import { NULL_DATE } from './constants';
+
```
JonasVanHoof, line 3: Or should we move this to the date-manipulation util file? Was not sure.. would move the date-manipulation thing over here ⌛

app/components/mandatarissen/fractie-selector.js
```diff
@@ -31,12 +32,12 @@ export default class MandatenbeheerFractieSelectorComponent extends Component {
 
   async loadFracties() {
```
JonasVanHoof, line 33: Would've liked to refractor it but maybe not in this ticket (a task to call an async method to just know if the options are being loaded) . Solution  feels like a combination with a _tracked function_

app/utils/date.js
bfidlers: you can certainly do that, would put the contents in the `date-manipulation` file in this one.

bfidlers (dismissed):
Some issues:

The verplicht validations don't really seem to trigger if the field is already blank or if you go back.
Screenshot 2025-04-25 at 17 18 47

Also some validations show, but they don't really prevent you from doing anything, you can just ignore it and go to the second page of the form.

app/utils/date.js
JonasVanHoof: Would prefer the util to be name date.js instead of date-manipulation.js but will leave it for now

Rahien (changes requested):
you're being too nice making the date checks warnings. we should block continuing in the modal if the dates are incorrect

app/components/mandataris/edit/form.hbs
```diff
@@ -95,24 +75,39 @@
...
+        {{#unless (await this.hideFractieField)}}
+          <Mandatarissen::FractieSelector
+            @fractie={{this.fractie}}
+            @isRequired={{@isFractieRequired}}
+            @onSelect={{this.updateFractie}}
+            @bestuurseenheid={{@bestuurseenheid}}
+            @bestuursperiode={{@bestuursorgaanIT.heeftBestuursperiode}}
+            @mandataris={{@mandataris}}
+            {{! TODO this should be generalised somehow }}
+            @isUpdatingState={{true}}
```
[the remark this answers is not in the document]
JonasVanHoof: Hm I only changed the #if statement and moved the component a bit up but I think we can clean it up. Should we need it if we handle this update depending on the reason of editing? 

Thats tuff no? As we do not want the person to switch fractions . hmm I guess thats ok => will put it to false so they can always select every fraction. (was just thinking about the previous fractie business rules * that I can remember)

app/components/mandataris/edit/date-input.js
```diff
@@ -0,0 +1,149 @@
...
+
+import { INPUT_DEBOUNCE } from 'frontend-lmb/utils/constants';
+import {
+  endOfDay,
+  isDateInRange,
+  isValidDate,
+} from 'frontend-lmb/utils/date-manipulation';
+
+function getWarningMessages() {
+  return trackedFunction(async () => {
```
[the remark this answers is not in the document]
JonasVanHoof: You are absolutely right 👍🏻

app/components/mandataris/edit/form.hbs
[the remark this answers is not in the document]
JonasVanHoof: Done

tests/integration/components/mandataris/edit/start-end-date-test.js
```diff
@@ -0,0 +1,27 @@
...
+module(
+  'Integration | Component | mandataris/edit/start-end-date',
+  function (hooks) {
+    setupRenderingTest(hooks);
+
+    test('it renders', async function (assert) {
+      // Set any properties with this.set('myProperty', 'value');
+      // Handle any actions with this.set('myAction', function(val) { ... });
+
+      await render(hbs`<Mandataris::Edit::StartEndDate />`);
```
Rahien, line 15: since these are just auto-generated tests, i would remove them

Rahien (changes requested):
Something seems to be going wrong with the selected dates. when updating state, the dates are ignored and today's date/existing end date are used

Rahien (changes requested):
current date is still used as start date when updating state

PR #545, opened 2025-04-24
really disable custom orgs

Rahien:
## Description

really disable custom organen: no adding new organen, no showing them, no adding mandatarissen to them
## How to test

try and find/add a custom orgaan/mandataris with the custom-organen feature flag disabled. it should not be possible.


bfidlers (approved):
This disables most stuff. 
It is still possible to get to a custom organ if the already added mandates to it, you can the click on the person and from the mandate page of the person you can click through to the organ. 
It is possible to restrict this access with an afterModel on the orgaan page, but then of course we will still be able to see the custom mandates.

An example bestuur which has data in a custom orgaan is Gemeente aarschot http://localhost:4200/organen/af7f9890-dba7-4ef9-a3b1-093b49ee91e1/mandatarissen

PR #548, opened 2025-04-25
allow linking to single or multiple persons

Rahien:
## Description

Allows adding fields to link to person instances (single and multi)
image
image

## How to test

- go to a custom form, add a new field to add a single person
- try adding and removing the person
- go to a custom form, add a new field to add multiple persons
- try adding and removing persons
- check the required option
- view the linked person and check if the person links are shown for single and multi forms
- make sure none of this is possible if the feature flag is off

## Linked PRs
- requires form-content PR https://github.com/lblod/form-content-service/pull/78/files

### DIFF
--- a/app/components/rdf-input-fields/custom-person-input.hbs
+++ b/app/components/rdf-input-fields/custom-person-input.hbs
@@ -6,7 +6,7 @@
   @formStore={{@formStore}}
 >
   <RdfInputFields::UnrestrictedPersonSelector
-    @inTable={{true}}
+    @inTable={{this.hideLabel}}
     @field={{@field}}
     @form={{@form}}
     @formStore={{@formStore}}
@@ -14,7 +14,7 @@
     @sourceNode={{@sourceNode}}
     @forceShowErrors={{@forceShowErrors}}
     @cacheConditionals={{@cacheConditionals}}
-    @show={{@show}}
+    @show={{or @show this.isReadOnly}}
   />
 
 </RdfInputFields::CustomFieldWrapper>
--- a/app/components/rdf-input-fields/custom-person-input.hbs
+++ b/app/components/rdf-input-fields/custom-person-input.hbs
@@ -1,0 +1,20 @@
+<RdfInputFields::CustomFieldWrapper
+  @inline={{true}}
+  @field={{@field}}
+  @form={{@form}}
+  @graphs={{@graphs}}
+  @formStore={{@formStore}}
+>
+  <RdfInputFields::UnrestrictedPersonSelector
+    @inTable={{true}}
+    @field={{@field}}
+    @form={{@form}}
+    @formStore={{@formStore}}
+    @graphs={{@graphs}}
+    @sourceNode={{@sourceNode}}
+    @forceShowErrors={{@forceShowErrors}}
+    @cacheConditionals={{@cacheConditionals}}
+    @show={{@show}}
+  />
+
+</RdfInputFields::CustomFieldWrapper>
--- a/app/components/rdf-input-fields/custom-person-input.js
+++ b/app/components/rdf-input-fields/custom-person-input.js
@@ -1,0 +1,14 @@
+import Component from '@glimmer/component';
+import { consume } from 'ember-provide-consume-context';
+
+export default class RdfInputFieldsCustomPersonInputComponent extends Component {
+  @consume('form-state') formState;
+
+  get hideLabel() {
+    return !this.formState?.isReadOnly;
+  }
+
+  get isReadOnly() {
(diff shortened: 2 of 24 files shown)


app/services/custom-forms.js
```diff
@@ -50,6 +50,19 @@ export default class CustomFormsService extends Service {
...
 
+  async getInstanceUsage(instanceUri) {
+    const response = await fetch(
+      `${API.FORM_CONTENT_SERVICE}/custom-form/find-usage?instanceUri=${encodeURIComponent(instanceUri)}`
+    );
+    const jsonResponse = await response.json();
+
+    if (response.status !== STATUS_CODE.OK) {
+      console.error({ jsonResponse });
+    }
```
JonasVanHoof, line 61: Just thinking at the rework throwing errors ticket. Should we check here on `response.ok` already? Than we do not need there status codes anymore (overtime ofcourse ;p)
```suggestion
    if (!response.ok) {
      console.error({ jsonResponse });
    }
```

app/utils/register-form-fields.js
```diff
@@ -149,5 +155,13 @@ export const registerCustomFormFields = () => {
         'http://lblod.data.gift/display-types/lmb/custom-concept-scheme-multi-selector-input',
       edit: RdfInputFieldCustomConceptSchemeMultiSelectorInput,
     },
+    {
+      displayType: PERSON_CUSTOM_DISPLAY_TYPE,
```
JonasVanHoof, line 159: Q: Should we also use the constants here for the other custom fields?
Not that you should do this in this PR, just for future implementations

app/components/rdf-input-fields/unrestricted-person-multi-selector.hbs
```diff
@@ -0,0 +1,47 @@
...
+        @skin="secondary"
+        @hideText={{true}}
+        {{on "click" (fn this.removePerson selected)}}
+      >Verwijder</AuButton>
+    {{/unless}}
+  </div>
+{{else}}
+  <div class="au-u-margin-bottom au-u-muted">
+    Geen personen geselecteerd
+  </div>
```
JonasVanHoof, line 29: Should we also make it a <p> tag as the person selector or should we both make them muted? 

Should the delete be disabled when no person selected?

image
https://github.com/lblod/frontend-lokaal-mandatenbeheer/blob/master/app/components/person/list-item.hbs#L2

app/components/rdf-input-fields/custom-person-input.hbs
```diff
@@ -0,0 +1,20 @@
+<RdfInputFields::CustomFieldWrapper
+  @inline={{true}}
+  @field={{@field}}
+  @form={{@form}}
+  @graphs={{@graphs}}
+  @formStore={{@formStore}}
+>
+  <RdfInputFields::UnrestrictedPersonSelector
```
JonasVanHoof, line 8: Looks good!
image

JonasVanHoof: Should we add sorting to the list of persons?

JonasVanHoof (approved):
Smooth!

Some comments but works nicely 💪🏻 

I didn't see a ticket for making the uris be actual values. Is this something you added to the finish up **forms** ticket (analyse)? 
image

app/components/rdf-input-fields/custom-person-input.hbs
Rahien: added to form ux ticket

app/components/rdf-input-fields/unrestricted-person-multi-selector.hbs
Rahien: will fix the non-muted single select. i'd keep the remove button, else things will jump around too much and that would annoy me as a user.

app/utils/register-form-fields.js
Rahien: if we reuse them, we should i think. But most are not reused.

app/services/custom-forms.js
Rahien: will change this

app/components/rdf-input-fields/custom-person-input.js
```diff
@@ -0,0 +1,14 @@
...
+export default class RdfInputFieldsCustomPersonInputComponent extends Component {
+  @consume('form-state') formState;
+
+  get hideLabel() {
+    return !this.formState?.isReadOnly;
+  }
+
+  get isReadOnly() {
+    return this.formState?.isReadOnly;
+  }
```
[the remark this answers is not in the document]
Rahien: wanted to keep it separate because of their different reasons..

PR #557, opened 2025-05-07
show validation errors throughout the app

Rahien:
## Description
show validation errors throughout the app
image
 
image
image
image
image
image
image

## How to test

gent is a decent size commune with a few validation errors, trigger validations and navigate through the app to see them. correct the validation errors and see them disappear when you refresh the validations (should take only a few seconds)

## Related PRs
requires https://github.com/lblod/app-lokaal-mandatenbeheer/pull/429

### DIFF
--- a/app/components/refresh-validations-button.hbs
+++ b/app/components/refresh-validations-button.hbs
@@ -2,13 +2,19 @@
   <div
     class="report-refresh-button au-u-flex {{if (eq @size 'small') 'small'}}"
   >
-    <AuPill @icon="clock-rewind" @size={{@size}}>Laatste validatie
-      {{this.lastRefreshedDate}}</AuPill>
+    <AuLink
+      @icon="clock-rewind"
+      @size={{@size}}
+      @route="report"
+      @skin="button-naked"
+      class="refresh-button-left"
+    >Laatste validatie
+      {{this.lastRefreshedDate}}</AuLink>
     <AuButton
       @disabled={{this.isGeneratingReport}}
       @skin="naked"
       @size={{@size}}
-      class="refresh-button-next-to-pill"
+      class="refresh-button-right"
       {{on "click" this.refresh}}
     >
       <AuIcon
--- a/app/components/refresh-validations-button.hbs
+++ b/app/components/refresh-validations-button.hbs
@@ -1,25 +1,31 @@
 <div class="au-u-flex au-u-flex--column">
-  <div class="report-refresh-button au-u-flex">
-    <AuPill @icon="clock-rewind">Laatste sync
+  <div
+    class="report-refresh-button au-u-flex {{if (eq @size 'small') 'small'}}"
+  >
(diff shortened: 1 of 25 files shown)


bfidlers:
The "Verkeerd aantal mandatarissen" on a specific date gives the wrong date.
Screenshot 2025-05-09 at 10 38 04

bfidlers:
Some validations are clear, but it is not clear how to fix it.
Screenshot 2025-05-09 at 10 48 17

app/components/refresh-validations-button.hbs
```diff
@@ -1,25 +1,37 @@
...
-    validatie fouten?</p>
-  <AuLinkExternal
-    class="au-u-flex au-u-flex-self-end"
-    href="https://abb-vlaanderen.gitbook.io/handleiding-lokaal-mandatenbeheer"
-  >
-    Bekijk de handleiding
-  </AuLinkExternal>
+  {{#unless @noMessage}}
+    <p class="au-u-flex au-u-flex-self-end au-u-padding-top-small">Meer weten
+      over validatie fouten?</p>
```
bfidlers, line 29:
```suggestion
      over validatiefouten?</p>
```

bfidlers:
I wonder if we want to link the report page somewhere permanently. You have the validatiefouten banner when validation errors were found, but the first time I go to a page it does not show (probably because it is still running in the background). Also if I have no validation errors there is no way of going to that page except if you go to the hard url. Maybe we should always show a banner that states if the validations are running, no errors were found or how many were found. Would maybe also add breadcrumbs and navigation in menu's.

bfidlers:
Going to the report page for bestuurseenheden with a lot of validation erros takes very long. It feels like the page is not responding, might want to add some loaders?

bfidlers:
I think we should also try removing duplicate validation errors
Screenshot 2025-05-09 at 10 58 23

bfidlers:
Sometimes when a validation ran there is a quick view that says everything is okay, before the actual errors are shown.

bfidlers:
It seems like not all validation errors are shown. I guess there is a limit in the call, but there is no way of paginating ...

bfidlers:
I think it would also be nice to be able to filter validations (in the report page) - but this is certainly not for this story, I just thought I should mention it.

bfidlers (changes requested):
Overal it looks pretty good, the validations are shown in different places which is nice, but the validations are not always clear, or it isn't always clear how you can fix them. And there are a few issues which do need fixing.

bfidlers:
This is a bit weird as well (for Antwerpen) 
Screenshot 2025-05-09 at 11 43 51

bfidlers (approved):
I still get an error when trying to go to a `mandatarissen.persoon.index` page, but don't think this is due to this story.

Screenshot 2025-05-09 at 16 08 05

Otherwise looks good!

PR #558, opened 2025-05-09
Ben/lmb 1465 start date replacement

bfidlers:
## Description

Add a modal when adding a replacement in the mandataris card. This should give the option to add a start date, fraction .. if the replacement does not exist already.

## How to test

Put a mandataris on verhinderd without selecting a replacement. When the mandataris is verhinderd add a replacement in the mandataris card. Both try adding a replacement that already exists (the person already has that mandate), and try adding one that does not exist, this behaviour should be different.


Rahien:

PR #560, opened 2025-05-13
custom shacl messages, fix css issues, hide feature flags in prod

Rahien:
## Description

Better shacl messages, fixes to scrolling, hide console.log in prod

image

(the message is still technical but this is an issue that should not happen if you use the application, people who complain can ask the helpdesk and get to us)

scrolling to limit the amount of screen space lost
image

## How to test

aalst and nieuwpoort on test are good to have a look at. also check the scrolling in the report page of nieuwpoort on wide, but not that high screens


JonasVanHoof (approved):
Looks good.

Not sure if the way its shown in a table on the mandataris page is the best

JonasVanHoof:
To issues:
 - frontend breaks when the report-generation service is not running
image

 - when I crop my view to be half my 24" screen the button has an overflow
image

PR #561, opened 2025-05-14
validatie improvements after final check

Rahien:
## Description

- no shouty banner on homescreen for validaties, use tag on top of menu
- add sidebar to validaties screen
- auto trigger validaties 20s after any put/post/patch except one that triggers validaties
- add info on validatie resultaten page
## How to test

- go to a municipality (e.g. gent)
- remove a replacement for a mandataris
- wait 25 seconds
- see that a validation error is now found and shown
- add the replacement again
- wait 25 seconds
- the validation error is now gone again
- you can do the same and trigger manual validations while waiting, you should see the auto validation being queued after yours have finished

### DIFF
--- a/app/components/shared/main-menu.hbs
+++ b/app/components/shared/main-menu.hbs
@@ -42,9 +42,32 @@
       {{/if}}
       {{#if (is-feature-enabled "shacl-report")}}
         <li class="au-c-list-navigation__item">
+
           <AuNavigationLink @route="report">
-            <AuIcon @icon={{this.validationIcon}} @alignment="left" />
-            {{this.validationText}}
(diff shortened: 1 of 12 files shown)


app/components/shared/main-menu.hbs
bfidlers, line 1: The UI bothers me a bit when the main menu is on top if your window is not big enough, when it is on the side it looks good.
Screenshot 2025-05-15 at 11 52 06
Screenshot 2025-05-15 at 11 52 20

bfidlers (approved):
UI feels a bit wonky in the main menu if your screen is not big enough, otherwise looks good!

PR #565, opened 2025-05-20
fix beleidsdomeincode selector

Rahien:
## Description

fixes the beleidsdomeincode selector so that it filters the list again AND that it allows adding new codes again
## How to test

filter the list of codes, add a new code, search for that code after

also, open the form for mandataris extra info edit it, customize that form, open and edit it again.

### DIFF
--- a/app/components/rdf-input-fields/concept-selector.js
+++ b/app/components/rdf-input-fields/concept-selector.js
@@ -2,48 +2,16 @@ import InputFieldComponent from '@lblod/ember-submission-form-fields/components/
 
 import { action } from '@ember/object';
 import { guidFor } from '@ember/object/internals';
-import { tracked, cached } from '@glimmer/tracking';
+import { tracked } from '@glimmer/tracking';
 import { service } from '@ember/service';
 
 import { NamedNode, Literal } from 'rdflib';
 import { hasValidFieldOptions } from '@lblod/ember-submission-form-fields/utils/has-valid-field-options';
 import { task, timeout } from 'ember-concurrency';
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -97,7 +97,7 @@
     "ember-page-title": "^8.2.3",
     "ember-plausible": "^0.2.0",
     "ember-power-select": "^7.2.0",
-    "ember-power-select-with-create": "^1.0.0",
```
Rahien, line 100: bumped the version while i was at it

app/components/rdf-input-fields/concept-selector.js
```diff
@@ -2,48 +2,16 @@ import InputFieldComponent from '@lblod/ember-submission-form-fields/components/
...
 import { task, timeout } from 'ember-concurrency';
-import { trackedFunction } from 'reactiveweb/function';
-import { use } from 'ember-resources';
 
 import { SEARCH_TIMEOUT } from 'frontend-lmb/utils/constants';
 import { FIELD_OPTION } from 'frontend-lmb/rdf/namespaces';
 import { showWarningToast } from 'frontend-lmb/utils/toasts';
-
-function getOptions() {
-  return trackedFunction(async () => {
```
Rahien, line 19: no matter what i tried, i couldn't entangle this.searchData with the tracked function. It never updated when its value changed, so i just rewrote the tracked function out of it and actively ask for it. Imo this also makes it more clear.

JonasVanHoof (approved):
image
image

* Can add new beleidsdomeinen
* Search is fixed
* selection still works
* IV still working

PR #566, opened 2025-05-21
warn about gaps in the timeline

Rahien:
## Description

gaps in the timeline are now also highlighted.
image

## How to test

edit a mandataris to create gaps and overlaps in the timeline. Fix them after.


JonasVanHoof (approved):
Very nice 🎈

PR #567, opened 2025-05-23
LMB-1549 wizard rework fixes

bfidlers:
## Description

Fixes some issues with the mandataris rework.

## How to test

See if the mandataris edit modal works for both corrigeer fouten and update state and if adding the replacement works as expected.


Rahien:

PR #568, opened 2025-05-23
LMB-1465 replacement modal v2

bfidlers:
## Description

Remove the option to select replacements from the mandataris card, move this logic entirely to the edit mandataris wizard.

## How to test

Check if you can add replacements in the mandataris edit wizard, but no longer in the mandataris-card


Rahien:

PR #573, opened 2025-06-03
mandataris edit wizard rework

Rahien:
## Description

combination of the different prs that affect the mandataris wizard.

I also added a couple of things on top:
- treat component arguments as immutable
- show what changed
- show different reasons for the change OR show correct mistakes
- bundle mirroring to ocmw with wizard
- optimizations so the changes aren't super slow
- removing some safeties that are not necessary anymore

## How to test

- do different updates to mandataris instances (corrections/update state)
- check if the changes are sent to ocmw (or not if asked not to)
- check that it's not possible to send changes to ocmw in kraainem (and that it also doesn't happen accidentally)
- check that mirroring is not available in ocmw or province

## Links to other PR's
to be tested with the optimizations here:
- https://github.com/lblod/mandataris-service/pull/116
requires the changes in app here:
- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/439


app/components/mandataris/edit/wizard.hbs
```diff
@@ -68,6 +68,19 @@
           stap.
         </div>
 
+        {{#if this.newStateDateWarning}}
```
JonasVanHoof, line 71: Duidelijk 
image

JonasVanHoof (approved):

PR #576, opened 2025-06-04
allow selecting vervanger when a verhinder mandataris is being creatd

Rahien:
## Description

allow selecting vervanger when a verhinder mandataris is being created
also fix the forms being inconsistent when going forward and back through the steps
## How to test

create a mandataris, mark as verhinderd immediately on create, select a replacement

also edit mandatarissen with replacments, go forward and back through the forms and check the saved state

### DIFF
--- a/app/components/rdf-input-fields/mandataris-replacement-selector.js
+++ b/app/components/rdf-input-fields/mandataris-replacement-selector.js
@@ -53,6 +53,7 @@ export default class MandatarisReplacementSelector extends InputFieldComponent {
       }
 
       this.loadMandaat();
+      this.refreshPersonFromForm();
       this.checkIfShouldRender();
     };
     this.storeOptions.store.registerObserver(onFormUpdate);
@@ -83,7 +84,6 @@ export default class MandatarisReplacementSelector extends InputFieldComponent {
     return this.args.field?.label || 'Tijdelijk vervangen door';
   }
 
-  // todo need to watch store and refresh every time
   async loadMandaat() {
     const forkingStore = this.storeOptions.store;
     const mandaatUri = forkingStore.any(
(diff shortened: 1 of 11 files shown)


app/components/rdf-input-fields/mandataris-replacement-selector.js
```diff
@@ -104,6 +104,26 @@ export default class MandatarisReplacementSelector extends InputFieldComponent {
...
+
+    if (!persoonUri || this.mandatarisPerson?.uri === persoonUri) {
+      return;
+    }
+
+    this.mandatarisPerson = (
+      await this.store.query('persoon', {
+        'filter[:uri:]': persoonUri,
+      })
+    )[0];
```
JonasVanHoof, line 124: What if a user was added to the form that is not in the database 💭

JonasVanHoof (changes requested):
Created a new mandataris and directly put im as verhinderd. Selected a replacement and saved the form

image

The mandataris is set to verhinderd but as you can see the replacement mandataris was not added?

app/templates/mandatarissen/persoon/mandataris.hbs
```diff
@@ -90,16 +90,17 @@
 
 </div>
```
JonasVanHoof, line 91: The mandataris  card shows the vervanger content when correcting the mandataris by removing the replacement 
image

app/components/rdf-input-fields/mandataris-replacement-selector.js
```diff
@@ -144,24 +164,22 @@ export default class MandatarisReplacementSelector extends InputFieldComponent {
...
-    this.replacements = mandatarises;
+    this.replacement = mandataris;
 
     // Retrieve options in store
     const matches = triplesForPath(this.storeOptions, true).values;
 
     // Cleanup old value(s) in the store
     matches
-      .filter((m) => !mandatarises.find((opt) => m.value == opt.uri))
+      .filter((m) => ![mandataris].find((opt) => m.value == opt.uri))
```
JonasVanHoof, line 178: This feels strange 😄 
Can't you not just do ... or am i not thinking straight 🙈 
```suggestion
      .filter((m) => mandataris.uri !== m.value )

```

JonasVanHoof (approved):
Looks ok, one thing in the mandataris card that is visible but should not be but not something you touched so approving 🆗

PR #577, opened 2025-06-05
lmb-1593 | dissalow deleting the custom form when it has usage

JonasVanHoof:
## Description

We do not want the form to be removed when it is used somewhere.

## How to test

1. Create an custom form
2. Link to this form from an extension of other custom form
3. Try deleting the form from step 1
4. You should only be able to close the delete modal
5. When deleting an instance of the form it can be deleted when there are implementations BUT these will be removed from the implementations when done (existing code)

## Links to other PR's

- https://github.com/lblod/form-content-service/pull/85

## Attachments

image

### DIFF
--- a/app/controllers/custom-forms/instances/index.js
+++ b/app/controllers/custom-forms/instances/index.js
@@ -93,9 +93,16 @@ export default class CustomFormsInstancesIndexController extends Controller {
     this.formToDelete = null;
   }
 
+  get formUsageLabels() {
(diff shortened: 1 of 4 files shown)


bfidlers (approved):
Works as expected. But a few remarks

1. This messages is a bit confusing: What do you mean with `implementaties`? If you mean instances of the form then the message is wrong, it always seems to show 1 even if I have more. Also it is just not clear what is meant.
2. The buttons are so far apart.

Screenshot 2025-06-13 at 09 38 36

PR #580, opened 2025-06-10
LMB-1521 | check if the the rijksregister number is already used

JonasVanHoof:
## Description

We want to check if the creation of a person will lead to a duplicate rrn. Show a warning when the persons rrn is already in the current graph.

## How to test

1. GO to an OCMW and to BCSD
2. start the creation of a mandatris
3. search for a person and than click create a new person
4. the validation should have trigggered for duplicate when searching on the rrn

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/442

## Attachments

**LATEST**
image

**OLD**
image

### DIFF
--- a/app/components/rdf-input-fields/rijksregister-input.hbs
+++ b/app/components/rdf-input-fields/rijksregister-input.hbs
@@ -23,6 +23,6 @@
   <AuHelpText @error={{true}}>{{error.resultMessage}}</AuHelpText>
 {{/each}}
 
-{{#each this.warnings as |warning|}}
+{{#each this.allWarnings as |warning|}}
   <AuHelpText @warning={{true}}>{{warning.resultMessage}}</AuHelpText>
 {{/each}}
--- a/app/components/rdf-input-fields/rijksregister-input.hbs
+++ b/app/components/rdf-input-fields/rijksregister-input.hbs
@@ -23,6 +23,6 @@
   <AuHelpText @error={{true}}>{{error.resultMessage}}</AuHelpText>
 {{/each}}
 
-{{#each this.allWarnings as |warning|}}
+{{#each this.warnings as |warning|}}
   <AuHelpText @warning={{true}}>{{warning.resultMessage}}</AuHelpText>
 {{/each}}
--- a/app/components/rdf-input-fields/rijksregister-input.js
+++ b/app/components/rdf-input-fields/rijksregister-input.js
@@ -13,24 +13,24 @@ export default class RDFRijksRegisterInput extends InputFieldComponent {
 
   @service store;
   @tracked rijksregisternummer;
-  @tracked duplicateWarningMessage;
+  @tracked duplicateErrorMessage;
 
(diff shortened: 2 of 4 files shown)


app/components/rdf-input-fields/rijksregister-input.hbs
```diff
@@ -23,6 +23,6 @@
   <AuHelpText @error={{true}}>{{error.resultMessage}}</AuHelpText>
 {{/each}}
 
-{{#each this.warnings as |warning|}}
```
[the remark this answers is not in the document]
JonasVanHoof: Ticket said warn but will update :D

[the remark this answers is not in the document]
JonasVanHoof: It shows as an error now and is blocked for saving. It is only blocked as we added the RequiredConstraint to the field, so also added this in my check in the code when we check on duplicates.

app/utils/form-validations/register.js
bfidlers, line 1: It would be cleaner if you could use the registered validation. But I get why it is difficult to do this, regarding the store etc. But it really does not make sense to register the validation and then just not use it.

app/components/rdf-input-fields/rijksregister-input.js
```diff
@@ -13,18 +13,32 @@ export default class RDFRijksRegisterInput extends InputFieldComponent {
...
+      return [
+        {
+          resultMessage: this.duplicateErrorMessage,
+        },
+      ];
+    }
+
+    return super.errors;
+  }
+
```
bfidlers, line 35: Can't we use the logic of the InputfieldComponent for this, by just writing to the `errorValidations` or extending the `updateValidations` method in it's entirety?

bfidlers (changes requested):
Code feels a bit messy.

Also I tried it multiple times and it still allows me to add duplicate rrns between gemeente and ocmw (it does not show a warning or error or anything).

app/components/rdf-input-fields/rijksregister-input.js
```diff
@@ -35,14 +49,48 @@ export default class RDFRijksRegisterInput extends InputFieldComponent {
...
+  async setFormValue(rijksregisternummer) {
+    let formValue = rijksregisternummer ? rijksregisternummer : null;
+
+    await this.checkForDuplicates(this.rijksregisternummer);
+    if (this.isDuplicate) {
+      formValue = null;
+    }
+    replaceSingleFormValue(this.storeOptions, formValue);
```
JonasVanHoof, line 94: For me it is showing the error and blocking the form from saving 🤔  curious!

JonasVanHoof:
Closing as we we want to do this the cleaner way with an asynchronous validation. The feature for async validations is still ongoing but we are using this validation as it's usecase

https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/583

PR #581, opened 2025-06-10
LMB-1634 | standard field visualization in editor

JonasVanHoof:
## Description

Standard fields ar not selectable which is strange for the user.

- show in the side editor that it cannot be edited
- show a lock icon next to the field label
- add a tooltip to the icon  
- change `Bewerk formulier definitie` to `Bewerk formulier` (Checked with Tim)

## How to test

Open a form extension (person extra info or mandataris extra info) and see it works as expected

## Attachements

image


bfidlers (dismissed):
It only seems to work for me if the form has a field I added myself.

https://github.com/user-attachments/assets/74e039fd-20e9-41d6-844f-2a2bbe276d9d

Rahien (approved):

PR #584, opened 2025-06-13
lmb-1634: fix show text in sidebar when clicking on  a standard field

JonasVanHoof:
## Description

When clicking the standard field is should show a text that it is not editable. Previous code broke deleting a custom form. This is fixed.

## How to test

Test out the edit form of a form extension

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/581


bfidlers:
Whenever I try to test I get the following, but this is also in master, not sure yet where it goes wrong.

Screenshot 2025-06-17 at 08 53 25

bfidlers:
It seems to work, but the selecting of the standard field does not really show.

https://github.com/user-attachments/assets/a1cbfa5b-980b-4774-a718-d636e5dddac2

Also, could you make a PR to add a version tag and update this in the backend?

PR #585, opened 2025-06-18
Ben/lmb 1633 show confirmation window when making person independent

bfidlers:
## Description

When making a person independent a confirmation modal is shown.
This window also allows selecting a specific date when the change should happen.

Screenshot 2025-06-18 at 14 14 11

## How to test

Go to the mandaten page of a person and check if the "become onafhankelijk" shows a modal with an input field and if the different actions work as expected. 
Also check if the create mandataris and the beeindig still work as expected.

### DIFF
--- a/app/controllers/mandatarissen/persoon/mandaten.js
+++ b/app/controllers/mandatarissen/persoon/mandaten.js
@@ -25,6 +25,7 @@ export default class MandatarissenPersoonMandatenController extends Controller {
   @tracked size = 20;
   @tracked canBecomeOnafhankelijk;
   @tracked currentNonOnafhankelijkeMandatarissen = [];
+  @tracked isIndependentModalOpen = false;
   @tracked isModalOpen = false;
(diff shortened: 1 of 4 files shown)


app/controllers/mandatarissen/persoon/mandaten.js
```diff
@@ -104,21 +102,20 @@ export default class MandatarissenPersoonMandatenController extends Controller {
...
           fractie: onafhankelijkeFractie,
         }
       );
-      const newMandataris = await this.store.createRecord(
-        'mandataris',
-        newMandatarisProps
-      );
+      const newMandataris = await this.store.createRecord('mandataris', {
+        ...newMandatarisProps,
+        rangorde: mandataris.rangorde,
```
JonasVanHoof, line 117: Why don't you add the rangorde to the create props method?

JonasVanHoof (approved):
Looks good, also start of day

PR #588, opened 2025-06-26
retry safe requests with timeout 

Rahien:
## Description

retry safe requests with timeout to work around backend issue where patch requests inexplicably time out sometimes

## How to test

use the dev tools to simulate slow connections, you should see get requests retried 3 times after 3 seconds and patch requests after 1 second. Other requests will not be retried.

the case where this was an issue was the patch of a mandataris instance when correcting a start date. Try that one in particular.


JonasVanHoof (changes requested):
Not sure what to think 🙈  With no network throtling it works good! But when setting it to 3G speeds i get this one,
image

bfidlers (approved):
Works for me

PR #591, opened 2025-07-08
lmb-1675: make sure to use startOfDay when creating a mandataris

JonasVanHoof:
## Description

When a mandataris is created the start date should always be at the beginning of the day.

## How to test

1. Create a mandataris
2. Update/correct a mandataris
3. Test the test branch for the mandataris service

## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/120
- https://github.com/lblod/mandataris-service/pull/121

### DIFF
--- a/app/models/mandataris.js
+++ b/app/models/mandataris.js
@@ -3,6 +3,7 @@ import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
 import moment from 'moment';
 import { MANDATARIS_EDIT_FORM_ID } from 'frontend-lmb/utils/well-known-ids';
 import { JSON_API_TYPE } from 'frontend-lmb/utils/constants';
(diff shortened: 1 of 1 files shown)


app/models/mandataris.js
```diff
@@ -105,36 +106,55 @@ export default class MandatarisModel extends Model {
     return approvedStatussen.includes(this.publicationStatus?.get('uri'));
   }
 
+  // Why is this here?
   get displayEinde() {
     return this.einde;
   }
```
JonasVanHoof, line 112: Its used but i don't see why we should use this instead of this.einde

Rahien: used to be more complex, when end date was 00 of the next day, but the GN team requested this to be changed. I left the getter in case we ever needed to do something fancy again (and because it was less of a hassle to change all its usages)

Rahien (approved):

PR #592, opened 2025-07-09
fix for safe retry timeouts breaking ember store sometimes

Rahien:
## Description

The standard AbortSignal.timeout approach is not good enough because it can cancel the request after it succeeds but before the body is read, which breaks ember functions as they don't expect this. This approach takes manual control over the abortion and clears the timeout once the request completes so the body can always be read without the timeout firing in between.
## How to test

use request throttling or set the timeout to a very short amount of time in the code


JonasVanHoof (approved):
image


Looks good 💪🏻

PR #594, opened 2025-07-10
LMB-1655 |  hide publication status for eigen organen

JonasVanHoof:
## Description

Eigen organen should not show publicatie status in mandataris card and overview mandatarisTbale.

## How to test

1. it is hidden in the mandataris card
2. the column is hidden in the mandataris table of an orgaan
3. extra fix when selecting a replacement and not filling in any fields and just going to the next step the values are passed on on exit of the component so a start date is set (first try it on master and see that it misses a start date)


Rahien (changes requested):
works, but can you remove the 'bekrachtig mandatarissen' option in the 'Meer acties' menu of custom organs too?

PR #599, opened 2025-07-18
lmb-1710: return the OG replacement when it did not change

JonasVanHoof:
## Description

The replacement is removed when correcting the end date of the mandataris.

## How to test

1. set status to verhinderd
2. select replacement for that mandataris UPDATE-STATE
3. Open the wizard again and change the end date
4. The replacement should still be added


Rahien (dismissed):

bfidlers (approved):
Works as expected.

PR #601, opened 2025-07-25
LMB-1691 | Hide fractie from bcsd

JonasVanHoof:
## Description

BCSD mnandatarissen should not have fracties. Hide the columns, values in the mandataris card when the mandataris or orgaan is BCSD.

## How to test

1. Check the mandatarissen for an OCWM BCSD orgaan
2. Fractie should be present in the mandataris card + tables (orgaan/mandatarissen + bulk bekrachtiging)
3. Run the migration from the APP 
4. All fracties should be removed and the fractie column should be hidden in the frontend 

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/458

### DIFF
--- a/app/routes/mandatarissen/persoon/mandataris.js
+++ b/app/routes/mandatarissen/persoon/mandataris.js
@@ -37,8 +37,10 @@ export default class MandatarissenPersoonMandatarisRoute extends Route {
       );
 
     const bestuursorganen = await mandaat.bevatIn;
-    const selectedBestuursperiode =
-      await bestuursorganen[0]?.heeftBestuursperiode;
+    const periodes = await Promise.all(
+      bestuursorganen.map((o) => o.heeftBestuursperiode)
+    );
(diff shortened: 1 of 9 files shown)


app/routes/mandatarissen/persoon/mandataris.js
```diff
@@ -94,6 +99,9 @@ export default class MandatarissenPersoonMandatarisRoute extends Route {
       bestuursorganen,
       firstBestuursorgaanInTijd,
```
JonasVanHoof, line 100: This needs to be fixed as bestuursorganen[0] is now breaking the the opening of a bcsd voorzitter

bfidlers (approved):
Bit worried about a few things:
1. We seem to be adding different checks in different places, sometimes we check politie organen, sometimes non polical organen ... Just feels a bit dangerous for creating inconsistencies
2. I'm not entirely sure where the fractie-selector gets hidden in the form, but I just want to double check, does this not create default memberships with the generators, or is this no concern?

PR #606, opened 2025-08-01
allow silencing validations

Rahien:
## Description

allow silencing validations
## How to test

go to aalst and mess with the data a little
then silence a couple of validations, check that they don't show up in the overview of the mandataris lists or history
then silence all
silence nothing
repeat checks


app/templates/report.hbs
```diff
@@ -11,12 +11,58 @@
...
+      <div class="au-u-box au-u-padding">
+        <div class="au-u-flex au-u-flex--between au-u-margin-bottom-small">
+          <AuHeading
+            @skin="5"
+            @level="2"
+            class="au-u-margin-bottom-tiny au-u-margin-bottom-none"
+          >{{this.validationErrorCount}}
+            actieve validatiefout{{if (eq this.validationErrorCount 1) "" "en"}}
+            gevonden</AuHeading>
+        </div>
```
JonasVanHoof, line 23: Should we add in `au-u-muted` text "X verborgen fouten" unde this statement of active errors

Or a number somewhere on how many hidden errors there are

app/components/validations/overview-callout.hbs
```diff
@@ -1,31 +0,0 @@
...
-    >{{this.validationErrorCount}}
-      validatiefout{{if (eq this.validationErrorCount 1) "" "en"}}
-      gevonden</AuHeading>
-  </div>
-  <div class="au-u-flex au-u-flex--start au-u-flex--spaced-tiny">
-    <p>Validatiefouten zijn mogelijke inconsistenties die we hebben gevonden in
-      de ingevulde gegevens. Deze validaties worden kort na elke aanpassing
-      uitgevoerd, maar je kan ze ook handmatig starten.
-    </p>
-
```
JonasVanHoof, line 16: This comment is not for this line of code but

I got this situation that 
1. I created a new mandataris
2. (now it should have a count missmatch) refreshed the validation in the card above the table
3. Got a duplicate rule

Image

JonasVanHoof: When refreshing it still had the two rule in the card

app/adapters/application.js
```diff
@@ -28,12 +28,15 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
         arguments[1] &&
         arguments[1].method != 'GET' &&
         [
```
JonasVanHoof, line 30: When i resolved a validation error (no replacement)  by adding a replacement to the mandataris => the validation rule check was not updated?

app/templates/report.hbs
Rahien: it's on the page itself if there are no items in the list of active validation errors but there are some hidden ones. the idea of hiding them is that they aren't in your face all the time.

PR #607, opened 2025-08-01
make orgaan clickable in mandataris

Rahien:
## Description

make the orgaan name clickable in mandataris
image

## How to test

go to a mandataris, visit their orgaan from there


JonasVanHoof (approved):
Nice addition!

PR #609, opened 2025-08-07
force selection of data for rangorde

Rahien:
## Description

force users to select a date before changing the rangorde of mandatarissen. they don't read the text otherwise.

this also changes the date input so it triggers an onchange if the input is the right length (without placeholders)
## How to test

change rangord
go to another route
go to the rangorde route again, the date should be empty again (but it will be filled if you use the back functionality of the browser, which is fine)
also check if you can still update a date elsewhere in the application

### DIFF
--- a/app/components/date-input.hbs
+++ b/app/components/date-input.hbs
@@ -22,6 +22,7 @@
       placeholder="dd-mm-yyyy"
       {{au-inputmask options=(hash mask="99-99-9999" placeholder="_")}}
       {{on "blur" (perform this.onChange)}}
+      {{on "input" this.onInput}}
       @disabled={{@disabled}}
     />
   {{/if}}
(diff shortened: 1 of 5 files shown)


app/components/date-input.hbs
```diff
@@ -22,6 +22,7 @@
       placeholder="dd-mm-yyyy"
       {{au-inputmask options=(hash mask="99-99-9999" placeholder="_")}}
       {{on "blur" (perform this.onChange)}}
+      {{on "input" this.onInput}}
```
JonasVanHoof, line 25: Didn't know this was allowed!

JonasVanHoof (approved):
Hard block for editing the rangorde without the date set, just what they needed.

Only question is if it needs a @from date that is today or `today - day`

PR #626, opened 2026-03-23
enable rangorde for districtsraadsleden

Rahien:
## Description

enables rangorde for districtsraadsleden and enabled updating rangorde for those organs (e.g. borgerhout)

## How to test

create and edit a districtsraadslid with rangorde, update rangorde in bulk for that organ


JonasVanHoof (approved):
You are kicking ass

Image

PR #632, opened 2026-06-18
add legcay-inspector-support

Rahien:
revive ember inspector by adding legacy inspector support.

Test by going through app and using the ember dev tools, I can use the data tab again!


JonasVanHoof (approved):
Fantastic!

PR #636, opened 2026-06-29
LMB-1914 | Can search over all legislatures

JonasVanHoof:
## Description

The user needs an option to search mandates over all legislatures. Now this is limited to only one period or overige. Add an option ALLE. 

## How to test

1. On mandatarissen page the bestuursperiode "Alle" is added to the dropdown
2. When selecting this option we see mandates over all legislatures 
3. Beheer fracties in organen route will not show this option
4. Organen page bestuursperiode alle is not shown here

## Notes

- created concurrency tasks for the setup methods that do not need to be awaited on loading of the app 

## Links to other PR's

- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/503
- https://github.com/lblod/mandataris-service/pull/140

## Attachments

image


app/routes/application.js
```diff
@@ -24,12 +24,10 @@ export default class ApplicationRoute extends Route {
   async beforeModel() {
     await this.session.setup();
     await this._loadCurrentSession();
-    await this.formReplacements.setup();
-    await Promise.all([
```
Rahien, line 28: isn't it dangerous to replace this by perform calls? I don't think that waits for them to complete before running the actual models of application and the route the user ends up in, so if they refresh one of the routes using these services, i think there is a chance they are not properly set up.

JonasVanHoof: That was my goal that it didn't get awaited and just resolved in the background! Didn't have any issues with it but it is indeed possible. Would the better thing to do then when we notice things missing that we check if the setup is still running and postone until done?

app/components/download-mandatarissen-from-table.js
```diff
@@ -6,13 +6,19 @@ import {
...
 
   get downloadLink() {
+    let bestuursperiodeId = this.args.bestuursperiode?.id;
+    if (bestuursperiodeId === FAKE_ALLE_BESTUURSPERIODE_ID) {
+      bestuursperiodeId = null;
+    }
+
     return this.mandatarisApi.getDownLoadUrl({
-      bestuursperiodeId: this.args.bestuursperiode?.id,
+      bestuursperiodeId: bestuursperiodeId,
```
Rahien, line 21: fractieIds doesn't seem to be sent to the backend anymore

Rahien (changes requested):

app/components/download-mandatarissen-from-table.js
JonasVanHoof: done! was a missed transformation of my abstract array

Rahien (approved):

